PR #12, opened 2025-03-28
more informative id duplication error

### LINKED ISSUE #8: Give more informative error message wheb duplicated ids in dock_view
At the moment we simply inform the user, that ids are duplicated. We would like to refine this, so that the user knows which ids are duplicated.

nelstevens:
Closes #8


DivadNojnarg (commented):
Looks good. Could you add some unit test to it? Thanks!

PR #16, opened 2025-04-02
Implement move_panel + fix proxy issue

### LINKED ISSUE #15: Proxy ID should be prefixed in all proxy methods
So that we can create multiple dock_view instances without conflicts...

DivadNojnarg:
Fix #15 and #6.


DivadNojnarg:
@nelstevens Feel free to check this one (during the day 🤣).

R/add_panel.R
```diff
@@ -5,8 +5,13 @@
...
-  session$sendCustomMessage("add-panel", panel)
+add_panel <- function(
+  proxy,
+  panel,
+  ...,
+  session = shiny::getDefaultReactiveDomain()
+) {
+  if (panel$id %in% list_panels(proxy, session))
+    stop("The panel ID you used is already in use!")
+  session$sendCustomMessage(sprintf("%s_add-panel", proxy), panel)
```
DivadNojnarg, line 16: @nelstevens I realised that we did not prefix the message id by the proxy id which means, having 2 dock_view instances in the same app, they would interfere since messages are intercepted in the same channel (by the same handler).

PR #18, opened 2025-04-04
Feature move_group

### LINKED ISSUE #14: Proxy method: move_group
We can move panels but we can also move entire groups:

```js
let groupTarget = api.getPanel(`${m.options.group}`)
              panel.api.group.api.moveTo({
                group: groupTarget.api.group,
                position: m.options.position,
              })
              return null;
```

nelstevens:
Closes #14


nelstevens:
@DivadNojnarg based on the Dockview documentation (https://dockview.dev/docs/core/groups/move and https://github.com/mathuo/dockview/blob/8a3a7e5063109cefda308bd870f2b1854dad7513/packages/dockview-core/src/api/dockviewGroupPanelApi.ts#L11) index is a valid parameter for moveTo when applied to a group. However, I could not find out what use index actually serves when moving a group. I tried setting different values etc. but it didn't seem to make any difference. Any ideas?

DivadNojnarg:
@nelstevens From my testing, `index` is only used when moving a single [panel](https://github.com/mathuo/dockview/blob/8a3a7e5063109cefda308bd870f2b1854dad7513/packages/dockview-core/src/__tests__/dockview/dockviewComponent.spec.ts#L6404). I had to look at the [tests](https://github.com/mathuo/dockview/blob/8a3a7e5063109cefda308bd870f2b1854dad7513/packages/dockview-core/src/__tests__/dockview/dockviewComponent.spec.ts#L230) to understand how it works. None of the test involving groups uses `index` so I can't really tell you. I think it's fine for now to remove/omit it. I could check with the dockeView devs if this is a mistake or it is really useful.

nelstevens:
@DivadNojnarg Thanks! I removed the index parameter. It looks like groups also have ids (see screenshot). So far, we have completely ignored group ids. Maybe for move_group it would make sense to select by group id as well? However, this would make the function more complicated and if the user never specifies a group id manually it might not even be useful. What do you think?
image

DivadNojnarg:
@nelstevens I see your point. We have `move_panel` that can move a panel within a group or in another group. This works based on panel ids.

It could be confusing for people that `move_group` works also with panel ids. You would expect to have something like:

```r
move_group <- function(id, target, position) {
}
```
where `id` and `target` refer to group ids (and not panels).

On the other hand, it's more work for users to think about group ids but if we document it well, it may work. I also don't expect users to use move_group more than move_panel.

With our current approach, user provide panel ids and our code is able to retrieve the corresponding group (as a side note, we have no way yet (from R) to check that 2 panels belong to 2 different groups so you could try to do invalid operations)

I'd be more in favor of renaming our current `move_group` to `move_group2` (saying that this works on the panel point of view and groups are implicitly inferred) and add a real `move_group` to consider group ids explicitly as shown above. 

How does this sound to you? We offer 2 APIs which avoids to have complex logic within a single function. On the JS side, we'd also have another handler for this.

nelstevens:
@DivadNojnarg Sounds great!

nelstevens:
@DivadNojnarg As a first step I renamed move_group to move_group2. I also changed some parameter names to something I view as more descriptive. However, I think it's still a bit difficult to understand. Maybe you have a better idea for parameter names? I'll wait with any further adjustments until we know what parameter names we prefer. Then I'll change them accordingly (also for add_panel etc.). Thanks!
Note: tests are broken currently. I'll fix later.

R/move_group.R
```diff
@@ -0,0 +1,37 @@
...
+#' move_group moves a group to a different position from withing a shiny server function. 
+#' The parameter source refers to the group-id you want to be moved. 
+#' Likewise destination refers to the group-id of a group you want to 
+#' select as destination.
+#' The difference between move_group2 and move_group is that move_group2 selects both 
+#' source and destination by panel-id, whereas move_group selects by group-id.
+#' @export
+move_group <- function(
+    proxy,
+    source,
```
DivadNojnarg, line 19: you could also use from and to if you need shorter parameters

DivadNojnarg:
@nelstevens I fixed test issues in #24. Don't worry to much with the tests here I can handle the conflicts when you are done with the feature.

DivadNojnarg:
Thanks @nelstevens. For now this is fine. We can fine tune details afterwards.

PR #19, opened 2025-04-12
Feature: Tab button

### LINKED ISSUE #13: Add tab button
Within a group, add a `+` button to spawn an empty tab. Users can then target it with `insertUI` to dynamically add content to it.

nelstevens:
Closes #13


DivadNojnarg:
Thanks @nelstevens. Will start the review on Monday when I am back from holidays.

nelstevens:
No rush! Enjoy your holidays :)

DivadNojnarg:
@nelstevens question:
- If the panel id is randomly generated with `e.timeStamp`, how do people add content to it from the Shiny server? We could probably expose it or write some placeholder within the panel body to say `replace me by insertUI(id = "#<RAND_ID>", ui = "New UI")` ... What do you think?

Observation: If I click on plus on the first group, then if I move the new panel to the left of the group, clicking on plus still adds a new panel to the old group. There's probably a missing piece on the JS side.

nelstevens:
@DivadNojnarg I've added a default panel content with instructions for insertUI. Let me know what you think. I wasn't able to reproduce your observation but encountered a similar bug (+ button did not work any more when dragging the new Panel out of its group to form a new group). Can you confirm that your observation is also fixed? Thanks!

DivadNojnarg:
I noticed `e.timestamp` would create invalid ids with `.` punctuation inside. We can use `Date.now()` instead. I also added an example to show how to replace content.

DivadNojnarg:
Test failing due to random ids.. Will fix later ...

Issue #20, opened 2025-04-14
Issues when ids aren't character


DivadNojnarg:
Seems that we can't do:

```r
library(shiny)
library(bslib)
library(visNetwork)
library(dockViewR)

nodes <- data.frame(id = 1:3)
edges <- data.frame(from = c(1, 2), to = c(1, 3))

ui <- page_fillable(
  dock_viewOutput("dock")
)

server <- function(input, output, session) {
  output$dock <- renderDock_view({
    dock_view(
      panels = list(
        panel(
          id = 1,
          title = "Panel 1",
          content = tagList(
            sliderInput(
              "obs",
              "Number of observations:",
              min = 0,
              max = 1000,
              value = 500
            ),
            plotOutput("distPlot")
          )
        ),
        panel(
          id = 2,
          title = "Panel 2",
          content = tagList(
            visNetworkOutput("network")
          ),
          position = list(
            referencePanel = 1,
            direction = "right"
          ),
          minimumWidth = 500
        ),
        panel(
          id = 3,
          title = "Panel 3",
          content = tagList(
            selectInput(
              "variable",
              "Variable:",
              c("Cylinders" = "cyl", "Transmission" = "am", "Gears" = "gear")
            ),
            tableOutput("data")
          ),
          position = list(
            referencePanel = 2,
            direction = "below"
          )
        )
      ),
      theme = "replit"
    )
  })

  output$distPlot <- renderPlot({
    req(input$obs)
    hist(rnorm(input$obs))
  })

  output$network <- renderVisNetwork({
    visNetwork(nodes, edges, width = "100%")
  })

  output$data <- renderTable(
    {
      mtcars[, c("mpg", input$variable), drop = FALSE]
    },
    rownames = TRUE
  )
}

shinyApp(ui, server)

```

PR #21, opened 2025-04-14
Fix #17 and #20

### LINKED ISSUE #17: Non valid direction argument in move_panel example?
Based on the [dockview documentation]( https://dockview.dev/docs/core/panels/add#positioning-the-panel ) direction: top does not seem to be a valid argument. However we use direction: top for our example here:
https://github.com/cynkra/dockViewR/blob/main/inst/examples/move_panel/app.R#L67

@DivadNojnarg am I missing something?

DivadNojnarg:
Will fix #17 and #20.


DivadNojnarg:
Feel free to check this one @nelstevens. It is quite big but overall this is about adding more checks, testing and also fixing 2 existing issues.

R/move_panel.R
```diff
@@ -5,7 +5,8 @@
 #' @param group ID of the panel you want to move the target to. They must belong
 #' to different groups.
 #' @param index Panel index. If panels belong to the same group, you can use index to move the target
-#' panel at the desired position.
+#' panel at the desired position. When group is left NULL, index must be passed and cannot exceed the
+#' total number of panel or be negative.
```
nelstevens, line 9: typo: total number of panel**s** ...

nelstevens (approved):
@DivadNojnarg Good stuff! :)

R/move_panel.R
DivadNojnarg: Thanks!

PR #24, opened 2025-04-17
F 22 serialise

### LINKED ISSUE #22: Feat: serialise/deserialize a dock
I discussed with @nbenn about implementing serialisation/deser of the dock: https://dockview.dev/docs/core/state/save, and this would probably allow us to keep track of everything in a single location (panels ids, group_ids, layout ...) and get rid of our current input to store panel ids.

This is useful for the blockr project.

### LINKED ISSUE #23: Bug: duplicated ids
It's currently possible to end up with panel with the same ids (in different docks). We should prefix the panel id by the dock id to ensure uniqueness.

DivadNojnarg:
Fix #22, Fix #23


DivadNojnarg:
Note: JS is totally broken in this branch. No idea why...

DivadNojnarg:
@nelstevens This PR provides a way to save the state and restore a dock from a saved state. Part of this state is also used to check panel ids, which replaces our custom input. I made also some adjustments here and there in the JS side by removing some events that are no longer needed. (Long run I'd like to get rid of jQuery).

There are some breaking changes (maybe not for your PR #18). We also fix an issue where panel id is now prefixed by the dock id to avoid duplicated html elements.

Feel free to check next week. One test fails but this is some random issue with chromote I can't replicate.

```
Caused by error:
! Chromote: timed out waiting for response to command Page.navigate
Backtrace:
```

DivadNojnarg:
Also fixed some issues with module to account for the namespace + renamed parameter and export the api object to make it accessible from JS.

PR #29, opened 2025-04-24
F vanilla

### LINKED ISSUE #26: Rmd/qmd: get rid of jQuery
Vanilla JS would avoid some errors when used without Shiny as jQuery isn't provided in that case.

nelstevens:
Closes #26


nelstevens:
@DivadNojnarg I couldn't find more jQuery occurrences that are not in shinymode. Maybe you'll find more.

DivadNojnarg (changes requested):
@nelstevens maybe we also want to remove jQuery from this: https://github.com/cynkra/dockViewR/pull/29/files#diff-1efbd341673c72ab94d40f3cebe8a222e97a9bca2528b2ab2c87862f047b843aR60 as it may also work without shiny.

PR #34, opened 2025-05-16
Feature: Split_view

nelstevens:


srcjs/widgets/splitview.js
```diff
@@ -84,18 +91,16 @@ HTMLWidgets.widget({
         api.onDidLayoutChange(() => {
```
nelstevens, line 91: @DivadNojnarg For dockview and gridview this eventhandler fires on first page load. However, it doesn't do this for splitview. Any ideas why that might be?

DivadNojnarg: Strange. It is documented as part of the API: https://dockview.dev/docs/api/splitview/api.

nelstevens:
@DivadNojnarg The way I see it, splitView does not support panel positioning (exp. position: { referencePanel: 'panel_4', direction: 'above' }). We could write a separate function for split_view panels or add some remarks in the panel function documentation + a check in split_view. What do you think would be better?

nelstevens:
> @nelstevens I had another look. It seems that the dividers are only visible when the widget is set to `orientation: horizontal`.



> @nelstevens I had another look. It seems that the dividers are only visible when the widget is set to `orientation: horizontal`.

@DivadNojnarg interesting. I can see the dividers in firefox but not in chrome. I'll check if I can find some css rules causing this.

PR #54, opened 2025-10-04
Refactor api

DivadNojnarg:
@nbenn: 

Still wip but you can test on your side...

- Now you benefit from `input$dock_initialized` to have correct timing.
- New `input[["<dock_ID>_restored"]]`.
- New `input[["<dock_ID>_panel-to-add"]]`.
- You can also initialize a dock with nothing by calling`dock_view()`.
- I introduced a proxy object for all `add_panel()`, ... methods as well as `get_dock()` and friends.
- New dock plugin API to control the add_tab and remove tab behaviors in a cleaner/safer way (instead of passing a basic list that had no semantic).
- `add_panel()` don't check for id uniqueness server side. This has been moved UI side.
- See news.md

- [x] Update some tests
- [x] Provide fine-grained inputs that are a subset of input$dock_state.


DivadNojnarg:
@nelstevens I think this PR is also a great place to remove `move_group`. Playing with this package for few months, I realise users are not supposed to know the group ids, as they are assigned by dockview. When you start to move things around, you can easily lose track of those groups. With the new API, we have a group/panel mapping available with the proxy data, but this is more meant to be used by internal functions rather than by users. So we can just use `move_group2` that works with panel IDs, where user have control over,  and rename it to `move_group`. Do you have any concerns?

PR #68, opened 2026-06-24
WIP: Debounce onDidLayoutChange to stop resize/bindAll storm

### LINKED ISSUE #67: onDidLayoutChange resize/bindAll storm pegs main thread on large boards (regression since dockview 4.10.0 renderer:'always')
## Summary

On boards with many resize-sensitive widgets (ECharts), `dockViewR`-driven apps
"load and load" / feel sluggish for a long time and the browser main thread is
pegged. Root cause is the `onDidLayoutChange` callback in
`srcjs/modules/callbacks.js`, which runs expensive work on **every** fire and
sets up a client-side feedback loop. This is a regression since the dockview
4.10.0 `renderer: 'always'` oscillation fix (shipped in dockViewR 0.3.0).

## The bug

`setDockViewCallbacks` registers:

```js
api.onDidLayoutChange(() => {
  window.dispatchEvent(new Event('resize'));   // → every htmlwidget on the page redraws
  if (HTMLWidgets.shinyMode) {
    saveDock(id, api);                          // serialize panel state to Shiny
    api.panels.map((p) => {                     // ...for EVERY panel:
      const pane = `#${id}-${p.id}`;
      Shiny.initializeInputs($(pane));
      Shiny.bindAll($(pane));                   // re-bind inputs/outputs
    });
  }
});
```

Since dockview 4.10.0 (the `#988` `renderer: 'always'` "Windows shaking" fix,
bundled in dockViewR 0.3.0), `onDidLayoutChange` fires **repeatedly while a
layout is still settling**. Each fire:

1. **dockview** dispatches a global `window.resize` — spuriously and repeatedly
   while the layout is unsettled. *(the trigger)*
2. **HTMLWidgets** catches every `window.resize` and calls `.resize()` on **every
   widget on the page** (not just the panel that changed). *(the amplifier)*
3. **ECharts** redraws — and the redraw nudges the layout → `onDidLayoutChange`
   fires again → **feedback loop**. *(the cost)*

The per-panel `Shiny.bindAll` is the dominant CPU cost per cycle (it re-scans
each panel's DOM). With `renderer: 'always'` every panel — including off-screen
ones — participates every cycle.

Plain tables / base-R plots never thrash because they are not resize-sensitive
and don't feed the loop. The chart block is innocent: it resizes via a
per-container `ResizeObserver`. The storm comes from HTMLWidgets' **global**
resize handler reacting to dockview's spurious dispatch.

## Measurements

Headless Chromium + Playwright; fps via `requestAnimationFrame`, `resize/s` via a
`window.resize` listener, blocked % via `PerformanceObserver('longtask')`. Rows
are stock dockview unless noted.

| Scenario | fps | resize/s | main-thread blocked |
|---|---:|---:|---:|
| real board (~99 panels, ~31 rendered ECharts) mid-thrash | **2** | **2.4** | **41 %** |
| same board, after a long settle | 60 | 0 | ~1 % |
| **same board + debounce fix** | **60** | **0** | **0 %** |
| raw dockViewR, 10 charts (no blockr.dock) | 60 | 0 | 0 % |
| 48 **table** blocks, 6 views | 60 | 0 | 0 % |
| 48 base-R **plot** blocks, 6 views | 60 | 0 | 0 % |
| 48 **ECharts** blocks, 6 views | 53 | 0 | 10 % |
| 66 ECharts, 11 views | 58 | 1.8 (transient) | 0 % |

The storm is **scale-dependent** with a threshold ~48 rendered ECharts; below
that it settles in one frame. Synthetic 144-chart repro storms at **38
resize/s**; with the fix it stays flat **60 fps, 0 resize/s** with all 144 charts
still rendered.

## Proposed fix

Wrap the `onDidLayoutChange` handler in a trailing-edge debounce (250 ms) so the
expensive work (resize dispatch + per-panel `bindAll` + state serialize) runs
**once, after the layout settles**, instead of on every intermediate frame.

Measured effect on the real board: **2 → 60 fps, 41 % → 0 % blocked**, all charts
still render, view-switching smooth. Synthetic 144-chart repro: **38 resize/s →
0**.

250 ms is a tuning knob (trades up to 250 ms delay on a genuine window resize for
killing the storm; could drop to ~150 ms if real resizes feel laggy). An
alternative worth considering is **scoping the resize to the panel that actually
changed** instead of a global `window.resize` dispatch.

A local fix branch implementing the debounce exists: `fix/debounce-layout-change`
(WIP PR to follow). Note the bundled `inst/htmlwidgets/dockview.js` still needs to
be rebuilt (`npm install && npm run production`).

## Regression origin

dockview `renderer: 'always'` oscillation fix — mathuo/dockview#988 — bundled in
dockViewR 0.3.0. Before 0.3.0 / dockview 4.10.0 the dock did not dispatch the
`onDidLayoutChange → window.resize` storm, so HTMLWidgets only resized charts on
real window resizes.

christophsax:
**WIP / draft** — debounces the `onDidLayoutChange` handler to stop the
resize/`bindAll` feedback loop that pegs the main thread on large boards.

Closes #67.

## What this does

`srcjs/modules/callbacks.js`: the `onDidLayoutChange` handler dispatched a global
`window.resize` + (Shiny mode) `saveDock` + `Shiny.bindAll` over **every** panel,
on **every** fire. Since dockview 4.10.0's `renderer: 'always'` oscillation fix
the callback fires repeatedly while a layout is still settling, and the global
resize → HTMLWidgets redraws every widget → ECharts redraw nudges the layout →
handler re-fires: a feedback loop. See #67 for the full analysis + measurements.

This wraps the handler in a 250 ms trailing-edge debounce so the expensive work
runs once after the layout settles.

Measured: real board **2 → 60 fps, 41 % → 0 % blocked**, all charts still render.
Synthetic 144-chart repro: **38 resize/s → 0**.

## Why it's still WIP

- [ ] **Rebuild the bundled widget** — only `srcjs/` source changed; the shipped
  `inst/htmlwidgets/dockview.js` is **not** rebuilt. Needs `npm install && npm run
  production` before merge.
- [ ] Decide the debounce interval (250 ms is a tuning knob — could drop to
  ~150 ms if real window-resizes feel laggy), or consider scoping the resize to
  the changed panel instead of a global `window.resize` dispatch.

## Notes

- Version bumped `0.3.0` → `0.3.0.9000`, NEWS entry added.
- Regression origin: mathuo/dockview#988 (`renderer: 'always'` oscillation fix),
  bundled in dockViewR 0.3.0.


christophsax:
Superseded by #69. The debounce here only **paces** the `onDidLayoutChange → resize → repeat` loop — it doesn't break it (the work still dispatches `resize`, which nudges the layout, which re-fires the handler; on a slow client it never converges). #69 keeps this debounce as its first commit and adds the two guards that actually terminate the loop (coarse size-signature dedup + a re-entrancy window). Closing this WIP in favour of #69.

PR #69, opened 2026-06-24
fix: break the onDidLayoutChange → resize feedback loop (not just debounce it)

christophsax:
Companion to BristolMyersSquibb/blockr.dock#252 (the "loads forever" reconcile loop). **Supersedes #68** — it contained only the debounce; this branch carries that commit plus the guards that actually break the loop.

## Problem

Even after the server-side reconcile loop is fixed, on a slow client (Safari, CPU contention) the dock keeps relaying out continuously: `onDidLayoutChange → global resize → widgets nudge layout → onDidLayoutChange → …`. Each cycle dispatches a global `resize` (every htmlwidget redraws) and re-binds every panel via `Shiny.bindAll`, re-rendering Shiny outputs server-side and pegging R at ~100% **at rest**. On boards with many resize-sensitive widgets (e.g. ECharts) it never converges.

A debounce (the original #68) only **paces** the loop — the expensive work still dispatches `resize`, which still nudges the layout, which re-fires the handler. It doesn't **break** it.

## Fix

`onDidLayoutChange` is debounced **and** guarded by two conditions before doing the expensive resize + rebind work (`onLayoutSettled` in `callbacks.js`):

1. **Coarse signature dedup** — skip when the layout structure + integer-rounded panel sizes are unchanged since we last acted, so the sub-pixel size drift the loop feeds on becomes a no-op and a pure echo terminates.
2. **Re-entrancy window** — for ~600ms after we act, ignore the `onDidLayoutChange` echo that our own `resize` dispatch provokes.

Genuine changes (real resize, add/remove, restore) still pass both guards.

## Validation

- Safari: R is **idle at rest** after the fix (was pegged at 100%).
- Chrome: view switches stay instant — no regression.

## Notes

- Supersedes #68 (debounce-only WIP), which is included here as the first commit.
- The remaining ~3-4s Safari view-switch latency is tracked separately in BristolMyersSquibb/blockr.dock#255 (output suspend / render cost), not this storm.


nbenn:
Pushed a commit (ed8390e) that swaps the loop-breaker from the timing window to a layout-tolerance compare. Same goal as your guards — break the `onDidLayoutChange → resize → onDidLayoutChange` feedback — but without the part that relies on a duration.

**Why it's needed.** The 600 ms re-entrancy window has two failure modes, and both bite exactly the slow / contended clients this PR targets. It's a fixed duration, so the echo from our own `resize` dispatch can arrive *after* the window closes and the loop resumes — the "echo outlasts the guard" problem, just at the 600 ms scale rather than the debounce's. And while it's open it drops *every* layout change, so a genuine user rearrangement that lands inside the window is silently never synced.

**What the commit does.** Keeps the debounce (burst coalescing) and the global `resize` dispatch (the only thing that actually re-fits widgets), and replaces `lastSig` / `suppressUntil` with `sameLayout(a, b, eps)`: the handler acts only when the layout differs from the one it last acted on by *structure* or a panel size beyond `RESIZE_EPS` (2 px).

**Why it's strictly better:**

- Timing-independent — a sub-pixel echo is a no-op no matter how late it arrives, so the slow-client case is handled by content rather than by a duration that can be outrun.
- Drops nothing — a real resize (>2 px) or any structural change always passes, so no user rearrangement is lost.
- Covers the boundary the rounding missed — a 1 px drift like `99.6 → 100.6` rounds to `100 ≠ 101` (so the old signature read it as a change and leaned on the window to swallow it), but `|99.6 − 100.6| = 1.0` is within tolerance.

**Validation.** The `sameLayout` logic is unit-tested (drift-is-same, real-resize-differs, structural-add-differs, focus-change-differs, and the exact integer-boundary case), and the bundle rebuilds byte-clean from `srcjs`. I could not reproduce the live loop locally — it needs a many-widget board under contention — so the real confirmation is against cedx, where you already have the repro and the CPU-throttle harness. If it holds there, the window comes out for good and the loop is broken by content, not timing.

I left the one-shot `onDidMaximizedGroupChange` resize dispatch alone — it's a discrete user action, not a loop driver.

nbenn:
One more CI failure the debounce introduces — separate from the snapshot churn, and a hard error rather than a snapshot diff: the `add_panel` test in `test-proxy.R` now errors with `Unable to find input binding for element with id dist`.

The cause is the debounce delaying `Shiny.bindAll`. The test does these two lines with no `wait_for_idle()` between them:

```r
app$click("btn")              # observeEvent -> add_panel() -> JS inserts the panel holding `dist`
app$set_inputs(dist = "norm") # dist isn't bound yet -> error
```

A newly inserted panel's inputs are bound by the `Shiny.bindAll` call inside `onLayoutSettled`, which is now debounced 250 ms. So the panel's DOM is in place but its `dist` input has not been bound when `set_inputs` runs, and the binding lookup fails. On `main` (no debounce) the bind ran on the same `onDidLayoutChange` tick, before the test's next step, so it passed.

Evidence it's the debounce specifically:

- New on this branch — `main`'s last R-CMD-check (May 12) was green on this exact test.
- Reproduces locally and in CI, and is present in the pre-`sameLayout` run too, so it's the debounce from #68, not the tolerance change.
- It aborts the `add_panel` test, so that test's later snapshots never regenerate — regenerating snapshots alone won't get CI green; the binding timing has to be fixed first.

The deeper point: binding a freshly added panel doesn't belong behind a *resize* debounce. Binding should follow panel insertion (e.g. `onDidAddPanel`), and only the resize / `saveDock` work — the part that's actually bursty — wants any coalescing. We're separately quantifying how much the debounce even buys on that side.

DivadNojnarg (approved):

PR #71, opened 2026-06-24
Tag _state updates with server/client provenance

### LINKED ISSUE #70: Suppress _state emission for programmatic layout changes
`onDidLayoutChange` sets the `_state` Shiny input on every layout change, including programmatic ones — `restore_dock` / `fromJSON`, `add_panel`, `remove_panel`. A Shiny app that mirrors `_state` back into server-side layout state therefore receives its own pushes as input: server push → `_state` echo → app folds it → server pushes again. A full restore emits a *cascade* of intermediate echoes (empty grid → partial → full) as the dock tears down and rebuilds, which on a slow / CPU-contended client sustains a feedback loop — panels flash, ~100% CPU on both server and browser, "loads forever". Downstream trace and reproduction: BristolMyersSquibb/blockr.dock#252.

The provenance needed to avoid this is already produced client-side: the `restore-state` handler fires `Shiny.setInputValue(id + "_restored", true, {priority: "event"})` right after `fromJSON`. So a programmatic relayout is already distinguishable from a user gesture at the source — `_state` just doesn't act on the distinction.

Proposal: don't emit `_state` for server-driven layout changes. A suppress flag set around the programmatic apply paths (the `restore-state`, `add-panel`, `remove-panel`, and move custom-message handlers) so the `onDidLayoutChange` handler skips `setInputValue` while it is up, then clears it — leaving genuine user interactions (drag, close, user-initiated move) emitting `_state` as today. Net effect: the server only ever sees user-initiated layout deltas and never has to recognise and filter its own echo.

Related: the resize-storm fix on `fix/break-resize-feedback` (#69) is the same theme — a programmatic relayout should not feed back. Worth solving together.

nbenn:
## Summary

The issue asks to stop `_state` from echoing server-driven layout changes back to a Shiny app that mirrors `_state` into its own layout state (the restore/reconcile loop in BristolMyersSquibb/blockr.dock#252). Outright suppression would break dockViewR's own contract: every exported `get_*()` reader (`get_panels_ids()`, `get_grid()`, …) reads `_state` and is documented for use right after a server op like `add_panel()`, so dropping those emissions makes the readers go stale. This reports **provenance** instead and leaves the filtering to the consumer.

- Add a companion input `<id>_state-source` (`"server"` / `"client"`), emitted alongside every `_state`, marking whether the change came from a server-side proxy call or a user gesture. `_state` itself is unchanged, so the read API, the restore round-trip, and every example app keep working — the new input is purely additive.
- Attribution is **causal, not time-based** (an earlier draft used a wall-clock window; this does not). dockview delivers `onDidLayoutChange` — where `_state` is emitted — on a microtask: its `AsapEvent` coalesces a burst of mutations into one notification fired *after* the api call returns. A flag raised around each proxy mutation is therefore cleared on a microtask queued right after the call; microtasks drain FIFO and dockview enqueues its notification *during* the mutation, ahead of that clear, so `saveDock` sees the flag raised for exactly the server-initiated change. No timers.
- The `onDidMovePanel` resize workaround fires a trailing `onDidLayoutChange` from a `setTimeout`, after the flag has cleared. It now carries the move's captured provenance forward, so a server-driven move stays `"server"` through that async tail rather than flipping to `"client"`.
- Every layout-mutating proxy handler is wrapped (`restore_dock`, `add_panel`, `remove_panel`, `move_panel`, `move_group`, `move_group2`, `select_panel`, `set_panel_title`). `save_dock()` (an explicit pull) and `update_dock_view()` (theme, not layout) are intentionally left to report the underlying state.
- Adds a `state_source` example and a shinytest2 regression test pinning the contract: `"server"` for add / move / restore, `"client"` for the initial render and a user-driven panel close.

The bundled `inst/htmlwidgets/dockview.js` is built against dockview-core 4.13.1 (now on `main`).

Fixes #70


nbenn:
Sorry @DivadNojnarg, I was too quick to mark this "ready for review". I first should try something out. Will ping you again when ready.

nbenn:
@DivadNojnarg The need for this change is no longer there in blockr.dock. It might be an interesting thing on its own. Or not. Your call. The idea here was to add information on whether a layout change was initiated client-side or is simply an echo from a server-side change. This would enable a down-stream consumer to swallow such "echo-triggers". Happy to discuss if you want to. But also fine to just close here. We can re-visit if need for such a thing comes up again.

DivadNojnarg (approved):
LGTM Thanks

nbenn:
Heads-up from #73 (#79, now merged to `main`): the approach changed during review from dropping the grid entirely to a structure-only `grid_shape()`, so the app-test grid snapshots changed shape on `main`. The exported `grid` is now topology only — branch / leaf nesting, orientation, group ids, panel membership, `activeView` — with every `size` / `width` / `height` / fraction dropped. Since this branch touches `test-dock_state.R` and its `dock_state` snapshots, they'll likely need regenerating against the new `main` when the stack rebases onto it. Flagging because the format change isn't obvious from a diff — comment only, I haven't touched the branch.

nbenn:
Heads-up on some churn I accidentally caused here. While doing post-merge cleanup after #83 (the `dockview-core` 4.13.1 bump) landed on `main`, I deleted its branch `81-dockview-413` as routine hygiene — without accounting for this PR still being stacked on it. GitHub auto-closed #71 as a result.

I immediately restored the branch to its exact prior tip (`6bfc14f`) and reopened this PR. The branch tip, diff, and history are all unchanged — nothing here was rewritten, only closed and reopened. Apologies for the notification noise.

Now that #83 is merged into `main`, the natural next step for this PR is to retarget it to `main` (a rebase will drop the already-merged bump commit, same as any destacking). I've left that to you rather than touching your branch.

PR #78, opened 2026-06-29
Add set_size proxy command to resize panel groups

### LINKED ISSUE #76: Add a proxy command to resize panel groups (proportional)
## Problem

The proxy command set covers add / remove / move / select panels and move groups, but there is no way to set a panel group's **size** from the server. Programmatic arrangement can place panels but not proportion them; the only current path to specific sizes is a wholesale `restore_dock` / `fromJSON` with sizes baked into the layout JSON, which tears down and rebuilds the whole dock.

## What to add

A `set_size` proxy command — the R proxy helper plus a `_set-size` custom message handler in `handlers.js` — that resizes a group through dockview's `panel.api.setSize`, parallel to the existing imperative handlers.

**It resizes a group, not a panel.** Groups (splitview cells) are dockview's only resizable unit: panels are tabs that fill their group, and the sashes sit between groups. The command takes a panel id for caller convenience (consistent with `move_panel` / `select_panel`) and resolves it to that panel's group, so resizing a panel resizes it and its tab-mates.

**The caller specifies a proportion; the wrapper converts to pixels.** The dockview method is pixel-based — `panel.api.setSize({ width?, height? })`, both in px — which is awkward from R, where the caller does not know the container width. So the command accepts a fraction along each axis (e.g. `width = 0.6`) and the wrapper converts using the dock's own dimensions (`api.width` / `api.height`) before calling `setSize`. The caller writes "60% wide", never a pixel count. v1 interprets the fraction against the **total dock** dimension, which is trivially queryable and deterministic; a fraction relative to the immediate parent split is a possible later refinement, as the parent splitview extent is not cleanly exposed on the public API. An explicit-pixel form can be added if a caller ever needs it.

## Independence

Self-contained, and a clear win on its own. It wraps a method the codebase already references (the `onDidMovePanel` workaround in `callbacks.js` calls `panel.api.setSize`), depends on nothing in the layout-ownership work, and composes for free: once #71 / #75 land, a server-issued `set_size` inherits the `"server"` provenance tag and the one-shot discrete emission like every other proxy op.

nbenn:
## Summary

- Adds a `set_size()` proxy command. The proxy set could place panels but not proportion them — the only path to specific sizes was a wholesale `restore_dock` that tears down and rebuilds the whole dock.
- Resizes the group a panel belongs to (groups are dockview's only resizable unit), taking a panel id for caller convenience like `move_panel()` / `select_panel()`.
- The caller passes a fraction per axis; the wrapper converts to dockview's pixel-based `setSize` against the dock's own `width` / `height`, interpreted relative to the total dock.

Fixes #76


DivadNojnarg (commented):
@nbenn Just one comment on this. If you press button 'skew' in demo and then press the 'resize', it seems that pressing resize again further changes the size of panel C.

See below:

Followed up on the resize oddity with headless measurements (chromote, default demo, `light-spaced` theme, widget width 942px).

**Symptom, quantified.** After skew + repeated identical `set_size(id = "C", size = 0.6)` clicks, C stays fixed while A and B drift on every click:

| step | A (px) | B (px) | C (px) |
|---|---|---|---|
| skew | 464.33 | 228.83 | 228.83 |
| resize 1 | 245.74 | 117.73 | 558.53 |
| resize 3 | 250.52 | 112.94 | 558.53 |
| resize 5 | 255.66 | 107.80 | 558.53 |

A:B goes from 2.09 to 2.37 over five clicks instead of holding the promised 2:1. Which panel appears to move depends on member order and viewport, so on my screen it was C.

**Root cause: mixed coordinate spaces.** `api.width` is 942 (includes the spaced theme's inter-group gaps), but the grid JSON leaf sizes are true view pixels: 307.33 x 3 = 922. `findSplit`'s `walk` rescales the JSON sizes to `rootWidth = api.width`, inflating every member by 942/922 (~2.2%). `setSize` then computes `extent`/`target`/`scale` in that inflated space but applies the results through `group.api.setSize()`, which operates in true view pixels. So:

1. Repeated calls are not idempotent: each click re-reads true sizes, re-inflates, re-applies; the error lands on the non-target members (~2.4px per click here).
2. The fraction is off even on the first call: C lands at 60.58% of the real extent, since `0.6 * 942` is computed against the gap-inclusive width.

**Suggested fix.** The `toJSON()` grid sizes are already absolute pixels in the same space `setSize` consumes, so `walk` should use them verbatim instead of rescaling to `api.width`/`api.height`. Then `extent = sum(members)` = 922, a repeated call yields `scale = 1` (exact no-op), and the requested fraction lands on the true splitview extent. With a gapless theme 942 = 922, which is why the bug is invisible there.

nbenn:
Fixed in the latest push — thanks, the mixed-coordinate-spaces diagnosis was exactly right, and reproducible.

One thing the headless measurements surfaced on top of it: `setSize()`'s *input* isn't in the toJSON content space either. It's a cell allocation that includes the cell's share of the split's gap. Measuring `setSize(A, X)` → `toJSON(A) = X − g` with `g` constant per call — ~5px for two groups in `light-spaced`, ~6.7px for three — i.e. `g = (api.width − Σsizes) / n`. So going verbatim (`extent = Σsizes`) fixes the *inflation* but each `setSize` still lands the view `g` short, and the sweep re-reads and re-applies that gap every click — which is the residual drift.

So the fix keeps your point — proportions are computed from the verbatim serialized (content) sizes — and additionally adds `g` back to every `setSize` input. The rendered extent needed for `g` is tracked down the grid, so a nested split uses its own gap rather than the root's.

After it, repeated identical `set_size(id = "C", size = 0.6)` on the demo is a true no-op, and C lands at exactly 0.6:

| step | A | B | C | C-frac |
|---|---|---|---|---|
| skew | 465.0 | 232.9 | 232.1 | 0.250 |
| resize ×1 | 247.9 | 124.1 | 558.0 | 0.600 |
| resize ×4 | 247.9 | 124.1 | 558.0 | 0.600 |

Verified idempotent for nested splits and the vertical axis too, and added an idempotency assertion to the e2e so it stays put.

DivadNojnarg:
@nbenn 

Follow-up after the gap-space fix: the fix itself holds (repeated identical `set_size` calls are now numerically exact no-ops, and the requested fraction lands on the true content extent). But alternating the demo's two buttons surfaces two further things, measured headlessly:

| step | A | B | C | A:B | C% |
| --- | --- | --- | --- | --- | --- |
| skew 1 | 461.00 | 230.50 | 230.50 | 2.00 | 25.00 |
| resize 1 | 245.87 | 122.93 | 553.20 | 2.00 | 60.00 |
| skew 2 | 461.00 | 93.33 | 367.67 | 4.94 | 39.88 |
| resize 2 | 306.70 | 93.33 | 521.96 | 3.29 | 56.61 |
| skew 3/4, resize 3/4 | identical to cycle 2 | | | | |

**Semantics (not a bug):** `skew` preserves the B:C ratio, `resize` preserves A:B. After resize 1, B:C is ~1:4.5, so skew 2 correctly splits the non-A half in that ratio and A:B lands near 5:1. The 2:1 in the button label only describes the demo's first path; alternating operations with different pivots cannot keep it. The demo labels overpromise.

**Real limitation:** dockview's group minimum width (100px default; read off the live groups) clamps the redistribution. At skew 2, B's computed cell allocation is 93.8px, below the minimum, so dockview clamps B to a 100px cell (93.33 content, exactly 100 minus its 6.67 gap share). The algorithm does not detect the clamp, so the leading-to-trailing `setSize` sequence dumps the deficit into the last member: C gets 56.61% instead of the requested 60%, and the preserved-ratio promise breaks (4.94 to 3.29).

Reassuring: cycles 2 through 4 are numerically identical to the hundredth of a pixel, so this is a stable fixed cycle, not creep.

Suggestion: make `set_size` constraint-aware: clamp each computed allocation to the member's `minimumWidth`/`maximumWidth`, redistribute the deficit among unclamped members, iterate until stable, so the target fraction is honored when feasible and degrades predictably when not. Alternatively document the best-effort behavior and soften the demo's button labels.

nbenn:
Thanks — agreed on both, and nice catch on the alternating path.

**Labels:** they did overpromise. The "2:1" only held on the first path; alternating `skew` and `resize` pivots on different members, so no fixed ratio survives. Softened both to "… share the rest, keeping their ratio", which holds whichever path you take.

**Min clamp:** a real limitation, and your account is exactly right — once a computed cell allocation drops below the group minimum, dockview clamps it and the leading-to-trailing sweep dumps the deficit into the last member, so the fraction is honored only when feasible.

On making it constraint-aware, the blocker is reading the live minimum: `group.api.minimumWidth` / `maximumWidth` come back `undefined` on the live groups here (they seem to be populated only on the constraint-change event), so a clamp-and-redistribute pass would have to assume dockview's 100px cell default rather than the actual per-group constraint. For now I've documented `set_size` as best-effort under the minimum (roxygen) and softened the demo. If you'd like the iterative clamp/redistribute anyway — against the 100px default, or via a live-min accessor I've missed — I'm happy to add it; just say which.

nbenn:
@DivadNojnarg — the constraint-aware question is a genuine design fork, so it's your call. Laying out the two options.

The relevant constraint: dockview doesn't expose the live per-group minimum. `group.api.minimumWidth` / `maximumWidth` read `undefined` on the live groups (they're populated only on the constraint-change event), so a clamp-and-redistribute pass can't read the actual minimum — it would have to assume dockview's 100px cell default.

**Option A — make `set_size` constraint-aware.** Clamp each computed allocation to the 100px default, redistribute the deficit among the unclamped members, and iterate to a fixed point. The requested fraction is then honored whenever feasible and degrades predictably instead of dumping the whole deficit into the last member. Cost: ~40 lines plus a test, and it bakes in the 100px assumption — which is silently wrong if anyone sets custom group constraints, since we can't read them back.

**Option B — best-effort (what's on the branch now).** Sizing is exact and idempotent while unconstrained; when a group would fall below its minimum, dockview clamps it and the fraction becomes approximate. This is now documented in `?set_size`, and the demo labels no longer overpromise.

For what it's worth, the tradeoff looks like it favours B to me — the clamp path is an edge case, the behaviour is stable (no creep), and A swaps a hard-coded 100px for a marginal-case gain. But you know the intended use better, so pick whichever you prefer and I'll implement it.

DivadNojnarg:
@nbenn We can go for B.

nbenn:
Pointer from #79 (dockViewR#73, now merged): while reviewing #79, @DivadNojnarg suggested a `set_size` idempotence check as a proportion assertion — apply `set_size(id, size)` twice and confirm the resulting fractions don't drift, with a tolerance. That belongs here rather than in #79's headless app snapshots. Headless renders never give the dock a real size (the container measures ~0 wide, so dockview emits degenerate / negative node sizes), which is exactly why #79 ended up structure-only — any size-derived value is unreliable there. With `set_size` you control the sizes, so an idempotence / tolerance assertion is actually meaningful. Flagging it here so it isn't lost in the merged #79 thread.

DivadNojnarg:
Heads-up for the rebase onto #80: with _state emission moving to gesture events, `set_size` (which surfaces only through `onDidLayoutChange`) will resize the layout while `_state` goes stale. Same pattern and fix as `set_panel_title` got in #80: `saveDock` inside the handler's `withServerDriven`.

nbenn:
Rebased onto main and retargeted — thanks for the heads-up, and it's handled (it actually let me simplify).

`set_size` persists explicitly inside `withServerDriven` (`setSize` + `saveDock`), the same pattern `set_panel_title` got, so `_state` reflects the resize tagged `"server"`. Confirmed against the merged emission model: on the demo, both `skew` and `resize` report `state-source = "server"`, and the e2e asserts it (plus the fraction lands exactly and repeats are idempotent).

On the shape, in case it's useful: an interim #80 state still had the one-shot `onDidLayoutChange` capture, which could re-emit a coalesced `"client"` flush a microtask after `withServerDriven`'s clear — I'd been carrying a macrotask-hold variant to cover that. The final #80 drops that one-shot, so a programmatic resize now hits no coalescer path at all; plain `withServerDriven` is enough, and I've removed the variant to match the convention. Green on all platforms.

PR #79, opened 2026-06-29
Snapshot grid structure in app tests, not pixel geometry

### LINKED ISSUE #73: Reduce reliance on pixel geometry in app-test snapshots
The app-level shinytest2 tests snapshot exported values with `app$expect_values(export = TRUE)`, which folds in `get_grid()`'s pixel geometry (`size`, `width`, `height`). That geometry is volatile and adds noise to the suite without testing anything these cases actually care about.

It bit twice while working on #70:

- The committed `remove_panel` / `select_panel` / `set_panel_title` snapshots were generated on an older render stack; the current CI runners render a few pixels differently, so every PR failed `R CMD check` on size-only diffs until the snapshots were rebuilt. One-time drift, but it recurs on each stack bump.
- `dock_state` (the `serialise` example) is genuinely flaky: the dock renders above a `reactjson` output that resizes, so its geometry lands on either side of a scrollbar boundary (568 vs 576 px) between runs. No committed value holds; #71 works around it by snapshotting structure only for that one test.

`app$expect_values(export = c(...))` already lets a test select specific exports, so dropping `grid` in favour of the structural facts these tests verify (panel / group ids, active panel / group / views, arrangement) is a small change that removes both the recurring drift maintenance and the `serialise` flake.

Filing so the geometry-in-snapshots noise is tracked rather than rediscovered on each stack bump, not because it's urgent.

nbenn:
## Summary

- The app-test snapshots folded in `get_grid()`'s absolute pixel geometry (`size`, `width`, `height`), which is volatile: recurring size-only `R CMD check` diffs on render-stack bumps, and a run-to-run flake in the serialise / `dock_state` test as a split landed on 568 vs 576 px. But the grid *tree* — nesting, orientation, group membership, order — is real coverage worth keeping; a dockview-core reshuffle or a layout regression should not sail through CI.
- Add `grid_shape()`: it returns the **structure** of `get_grid()` — branch / leaf nesting, orientation, group ids, panel membership and order — with all absolute geometry dropped. That is deterministic and still catches a structural or layout reshuffle, with no pixel flake. The ten example apps export `grid_shape(dock_proxy)` in place of raw `get_grid()`, and the 28 snapshots are regenerated. `get_grid()` is unchanged and stays exported.
- Snapshot structure only rather than normalized proportions — rationale in the comment below; short version, headless renders don't produce valid layout sizes, so a fraction can't be measured there honestly.
- A unit test pins the contract: structure and membership preserved, and no `size` / `width` / `height` / `fraction` survives.
- Wait for idle after `app$click("restore")` in the `dock_state` app test. The restore is an async client round-trip and the snapshot was being captured mid-restore, before the `test` panel re-appears. The `test-coverage` (covr) job surfaced it — its instrumentation loses the race that normal timing wins — and every other app test already waits after its action.
- Also removes a stray committed `dock_state-002.new.json`.

This re-adds a `grid` export that PR #71 (currently stacked on this branch) had adapted to the absence of, so #71's `dock_state` snapshots will need re-syncing once this lands.

Fixes #73


DivadNojnarg (commented):
@nbenn In another PR, where I started to bump dockView to v7 or v8 ... LOL ... I implemented a normaliser instead of totally removing the grid. The idea was to remove the flakiness but keep a way to detect upstreams (breaking) changes in dockview.

---

Here is Claude's summary:

The commit drops get_grid() from exportTestValues() across the example apps, so snapshots now pin only panel ids and the active group. What you lose is real: the grid tree (nesting, orientation, panel→group mapping, order) and any sizing behavior are now invisible to CI; a dockview-core bump that reshuffles layout JSON or a regression like the set_size drift we just diagnosed would sail through. But the flake was only ever the absolute pixels, so you don't have to choose between the two. Three mitigations, in the order I'd do them:

Re-export the grid through a normalizer instead of raw. Add a small test helper, say grid_shape(), that keeps the structural facts (branch/leaf nesting, orientation, group order, which panels live where) and replaces absolute size/width/height with sibling-relative fractions rounded to 2 decimals. The ±1-2px render-stack noise that caused the drift disappears at that rounding, while real behavior changes stay visible. Notably, the set_size non-idempotence bug (A drifting ~2.4px, 0.3% per click) would still show up after a few clicks in rounded ratios (2.09 → 2.37 ratio drift is huge at 2 decimals), so this recovers the class of bug you most recently cared about.

Property assertions with tolerance for the resize/move operations, alongside snapshots: after set_size(id = "C", size = 0.6), assert C / extent == 0.6 with tolerance = 0.01, and assert idempotence by applying it twice and comparing all fractions. Snapshots tell you something changed; these say what contract broke, and tolerances make them flake-proof by construction.

A dockview-core upgrade canary. The biggest breaking-change vector is bumping the vendored dockview-core (e.g. the 4.13.1 bump on the other branch). One unit test that snapshots the key-structure of toJSON() for a fixed two-panel dock (names and nesting only, zero numbers), plus an expect_identical on the bundled dockview-core version string, means any upstream format drift surfaces exactly at bump time as a conscious snapshot review rather than as downstream breakage in blockr.dock.

The one thing I'd not do is bring back raw pixel snapshots with a looser comparator; shinytest2's JSON snapshot comparison is exact-match, so tolerance has to live either in the exported values (rounding, option 1) or in explicit assertions (option 2).

nbenn:
@DivadNojnarg — I took the normalizer route first (re-export the grid with each `size` replaced by its sibling-relative fraction, rounded to 2 dp), and for most layouts it worked. But it fell over on one class of case, and understanding why is what pushed us to structure-only — writing it up here since it's the crux of the decision.

The normalizer produced sane fractions almost everywhere: of 105 fraction values across the 28 snapshots, 93 were valid — in `[0,1]` and summing to ~1 per split (`0.5/0.5`, `0.33/0.66`, `0.26/0.74`, …). The problem was the other 12, all in the `demo` layout, which came out `1.06` and `-0.06`. A fraction greater than 1 is obviously wrong, and the cause is that these tests run headless: the dock container measures ~0 wide, so when panel 2's `minimumWidth` (500px) can't be met, dockview-core hands its sibling a **negative** `size` (`-5`). `size / sum(siblings)` with a negative in that sum sails past 1.

The deeper point is that this isn't a rounding artifact I can guard away — it's that **headless renders don't produce valid layout sizes at all**. The container never gets real dimensions, so the pixel sizes aren't a meaningful partition to begin with. Pixels flake (the 568-vs-576 this issue is about), and fractions derived from them are either fine-by-luck or degenerate. There's no honest proportion to snapshot from these app tests.

So we snapshot the **structure** instead — branch/leaf nesting, orientation, group ids, panel membership, order — and drop every number. That's deterministic, and it still delivers what your review was protecting: the raw "drop grid entirely" lost the tree, and structure-only keeps it, so a dockview-core bump that reshuffles the layout JSON still surfaces as a snapshot diff. It just stops pretending to measure proportions where they can't be measured.

Proportions are still worth asserting — but where the sizes are real, i.e. a unit test with controlled input, not a headless app snapshot. Your mitigation #2 (the `set_size` idempotence check) is exactly that shape, and it belongs with `set_size` in #78 rather than here. Happy to add it there.

DivadNojnarg (approved):
@nbenn Thanks. Looks much better.

PR #80, opened 2026-06-29
Emit `_state` on discrete gesture boundaries, not per-frame layout changes

### LINKED ISSUE #75: Emit `_state` on discrete gesture boundaries, not per-frame layout changes
## Problem

The `_state` input is emitted from the `onDidLayoutChange` handler in `srcjs/modules/callbacks.js`, and that event fires *per frame*: during a sash drag dockview recomputes the layout continuously, re-fires while the layout is still settling after a structural change, and the global `resize` we dispatch to re-fit embedded widgets nudges them by a sub-pixel, which re-fires it again. That is the resize storm behind #67, and the `RESIZE_EPS` / `sameLayout` guard added in #72 exists *only* to filter it — a heuristic that absorbs sub-pixel echoes so the per-frame stream doesn't peg the main thread and flood Shiny with `_state` updates.

The per-frame emission is the actual defect; the debounce is symptom-treatment. A layout update is not a frame stream — it is a set of discrete user actions, each with a clean boundary.

## Fix

Trigger persistence on the discrete gesture events, not on the per-frame layout stream.

Every structural gesture already fires its own event exactly once — `onDidMovePanel` (move / reorder), `onDidAddPanel`, `onDidRemovePanel`, `onDidActivePanelChange`, `onDidAddGroup` / `onDidRemoveGroup`, `onDidMaximizedGroupChange`. Resize is the only continuous gesture, and it has a boundary too: the pointer-up that ends a sash drag. The library detects this internally (`onDidSashEnd`, fired from the sash pointer-up handler in `splitview.js`) but does not surface it on the public `DockviewApi`, so the wrapper hooks `pointerup` on the `.dv-sash` elements within the container.

Emit `_state` from those boundaries and the storm is gone at the source: one update per gesture, no sub-pixel echo to absorb. The `RESIZE_EPS` / `sameLayout` guard becomes deletable, and #72 is retired rather than carried forward.

## Split the two jobs

The `onLayoutChange` handler currently does two unrelated things in one place:

- `saveDock()` + the Shiny rebind — **state** emission. This moves to the discrete events.
- `window.dispatchEvent(new Event('resize'))` — re-fits embedded widgets (plots, ECharts, DT) into their new size. This is a **render** concern, not state.

The `sameLayout` guard currently tames both — the Shiny storm *and* a client-side loop where a widget re-fit nudges the layout, re-firing `onDidLayoutChange`, which re-fits again. Moving state off the per-frame path solves the Shiny side outright. The widget re-fit then becomes a separate, client-only decision — keep it live during a drag, or move it to sash-end as well — and must not be left riding the old per-frame path in a way that reintroduces its own feedback loop once the guard is gone.

## Correctness notes

- Container / window resize has no gesture boundary and is intentionally *not* persisted: a saved layout should carry proportions, not the pixel sizes of whatever window it last lived in. Keying off dock actions ignores environmental resizes by construction. An explicit `save_dock()` still reads `toJSON()` fresh, so a pull is always current.
- Discrete capture is exact for structure and sub-pixel for size — a structural change fires once, then sizes settle over a few frames, cosmetically identical on restore. Not worth chasing.
- The discrete set includes the server proxy ops (`add_panel`, `remove_panel`, `move_panel`, …), so the `get_*()` readers that read `_state` right after a server op stay fresh — no regression to that contract (cf. #70 / #71).

## Scope

This is independent of the controlled-component direction in #74 — it is correct under any ownership model, client-owned or server-owned, and should land ahead of that decision, since it removes the debounce machinery every later option would otherwise inherit.

### LINKED ISSUE #82: Drop the #1031 `setSize` workaround; coalesce `_state` on a microtask
Depends on #81 (the dockview bump).

PR #80's `_state` coalescer flushes on a `setTimeout(1)` — not by choice. The move's `setSize` (the #1031 workaround) has to run on a macrotask, and folding it into the single emission forces the flush to wait that long. That ~1 ms window is the one place two gestures can overlap: a server push and a user gesture landing in the same window coalesce into one emission tagged with whichever burst came first, so a genuine user change could read `"server"` and be ignored as the board's own echo (rare, self-healing on the next gesture, but real — flagged on #80).

Once dockview is at ≥ 4.13.1 (#81), the `setSize` workaround can go, and with it the macrotask requirement: the flush moves to a `queueMicrotask`. A microtask drains within the same task as the events that scheduled it, so a server op (a Shiny message handler) and a user gesture (a DOM event) are always in different tasks → different drains → separate flushes. Two gestures can no longer merge — the cross-burst window closes by construction rather than by a tag heuristic.

It's also a net simplification of `srcjs/modules/callbacks.js`: drop the `setSize` workaround, the `setTimeout`, and the provenance carry-forward (`serverDrivenFor` / `runServerDriven`). The live-flag read inside `saveDock` plus the before-clear microtask is all that's needed, and `onDidMovePanel` becomes a plain `requestSync` like every other event.

Prototyped locally (4.10.0 with the workaround removed, which is unsafe to ship there but fine to measure): one `_state` per gesture (`restore` 13 → 1), `test-state-source.R` 7/7, full suite green. Blocked only by the dockview version.

nbenn:
Fixes #75, fixes #82. Stacked on #71 (base `70-state-provenance`), which carries the #73 grid-drop and the #81 dockview bump to 4.13.1.

`input$<dock_id>_state` was emitted from the `onDidLayoutChange` handler, which fires per frame: dockview re-fires it while a layout settles, and the global `resize` dispatched to re-fit widgets nudges them a sub-pixel and re-fires it again. That per-frame stream is the resize storm; the `RESIZE_EPS` / `sameLayout` tolerance guard existed only to absorb its sub-pixel echo.

Emission now happens on the discrete gestures that each fire once — panel add / remove / move, group add / remove, active panel / group change, maximize, restore, and the pointer-up that ends a sash drag (hooked on `.dv-sash`, since dockview doesn't surface `onDidSashEnd`). Each event flags a pending flush, coalesced onto one `queueMicrotask`, so a compound gesture — a tab move that splits a group fires three events — emits a single settled `_state` instead of three (`restore`: 13 → 1). Consumers no longer debounce the burst (#82). The guard is deleted; widget re-fit moves off the per-frame path onto the same flush.

A container / window resize is the one layout change with no gesture boundary — and no event-driven end, since the window-drag-end belongs to the OS. But `_state` still has to reflect it: consumers read the input directly, to drive observers and to serialise the layout, so a stale `_state` after a resize is a bug. So it's debounced on a `ResizeObserver` over the container and persisted once it settles — `saveDock` only (widgets already took the browser's native `resize` live, so nothing is dispatched and the feedback loop can't reform). It reads `"client"` in `_state-source` — environmental, not server-initiated. The initial paint is captured separately by a one-shot (tagged `"server"`), which the observer seeds past without re-emitting.

Provenance (#71) is preserved with no extra machinery. The gesture flush is a microtask that drains in the same task as the events that scheduled it — inside the server op's `withServerDriven` window, ahead of its clear — so `saveDock` reads the correct `_state-source`. And because a microtask can't span tasks, a server op and a user gesture (always different tasks) never share a flush, so coalescing can't merge two differently-tagged gestures. `test-state-source.R` covers all of it, including the new resize case.

Dropping the `onDidMovePanel` `setSize` workaround is what lets the gesture flush sit on a microtask rather than a `setTimeout` — safe now that the base bundles dockview 4.13.1 (mathuo/dockview#1031, #81). `callbacks.js` sheds the `setSize` workaround, the `setTimeout`, and the provenance carry-forward (`serverDrivenFor` / `runServerDriven`); there are no snapshot changes — the diff is the code plus the new resize test.


nbenn:
**Follow-on within this PR's contract: coalesce `_state` to one emission per gesture.**

Moving the persist off the per-frame `onDidLayoutChange` onto discrete boundaries fixes the resize storm, but it leaves a gap: `persistState` (→ `saveDock`) is called synchronously inside *each* discrete handler, so a single *user* gesture that triggers several *dockview* events emits `_state` several times. The clearest case is a tab move that splits a group: `onDidAddGroup` and `onDidActivePanelChange` fire synchronously (two emits), then `onDidMovePanel`'s handler runs its `sync()` from the `setTimeout(1)` (a third).

The comment "No per-frame stream, so no debounce required" holds for the resize storm, but "not per-frame" isn't "once per gesture" — a compound gesture is still several events, so coalescing is still required *somewhere*. Right now that somewhere is the consumer: blockr.dock keeps a 250 ms debounce in its live-sync fold purely to collapse these bursts. It surfaced in BristolMyersSquibb/blockr.dock#274, where the board owns layout and each intermediate emission churns the board↔dock loop — dropping the debounce there immediately broke a panel-move test. That's a producer concern leaking into every consumer.

The contract this PR should hand downstream is **one settled `_state` per gesture**. Concretely: replace the per-event `sync()` calls with a single scheduled persist — any event flags "persist pending," one emit fires per burst carrying the final settled layout. The split you already made (`persistState` vs `refitWidgets`) keeps this small and on the persist side; the re-fit can stay per event if you like.

The one constraint to respect is provenance (#71): the server/client tag is read at emit time and is only valid inside the server op's microtask window — which is why `onDidLayoutChange` uses the `AsapEvent` microtask coalescer, and why the `onDidMovePanel` `setTimeout` carries the flag forward via `runServerDriven`. So the coalescer has to stay provenance-correct (microtask-scoped, or carrying the flag forward) — not a `requestAnimationFrame` / `setTimeout` that fires after the flag clears and would mis-tag a server push as "client."

Payoff: blockr.dock (and any other consumer) drops its debounce entirely, because the emitter honours "one per gesture" instead of the consumer guessing a settle time.

nbenn:
Done — pushed as one commit.

Each event now flags a pending flush instead of persisting; the first one records `serverDrivenFor(id)` synchronously inside the op's window and schedules the flush a tick out, past the synchronous burst and the move's `setSize` (which folds into it). The flush re-applies that provenance via `runServerDriven` after the op's microtask clear — so `prev` is already down and an in-flight op's flag can't be corrupted (the one-shot hazard), and the deferred emit stays correctly tagged. One settled `_state` per gesture: `restore` drops from 13 emits to 1, `add` 4→1, `move` 3→1, close 2→1; `test-state-source.R` stays green.

An unexpected bonus: with the persist coalesced into a single deferred `bindAll`, the hidden-tab slider stops binding — the same place the per-frame path left it. So the `obs` / `variable` shift these tests were picking up (and the closed-panel stale-input it surfaced) both go away; the input snapshots revert to this branch's base, and #80's diff is now code-only.

One boundary worth flagging: a user gesture and a server push within the same ~1 ms flush window coalesce into one emission carrying the *first* burst's tag. It's a sub-millisecond coincidence and the merged state is still the correct final layout, but a server push landing inside a user gesture's window would read `"client"`. Tightening it — flush-and-restart when the provenance changes mid-window — is more machinery; glad to add it if you'd rather not carry the edge.

nbenn:
Implemented #82 inline now that the base bundles 4.13.1 — thanks for the rebase.

Dropping the `setSize` workaround (dead weight once #1031 is fixed) removed the only reason the flush had to be a `setTimeout`, so it's now a `queueMicrotask`. That closes the cross-burst edge I flagged above at the source rather than with a tag heuristic: a microtask drains within the task that scheduled it, and a server op (a Shiny message handler) and a user gesture (a DOM event) are always separate tasks — so they can never land in the same flush. Two gestures can't merge into one mis-tagged emission, by construction.

It also sheds the provenance carry-forward: `callbacks.js` is back to just `import { saveDock }`, and `onDidMovePanel` is a plain `requestSync` like every other event. One `_state` per gesture (`restore` 13 → 1), `test-state-source.R` 7/7, full suite green, diff code-only. Updated the description.

nbenn:
Added the resize-settle path we discussed — you were right that a stale `_state` after a window resize is a real bug (consumers read the input to drive observers and to serialise, not `toJSON()`), and "intentionally not persisted" conflated what an explicit save bakes in with what the live input must reflect.

A container resize now goes through a debounced `ResizeObserver` on the container: one `saveDock` once it settles, tagged `"client"` (environmental, not server-initiated, as you called). It's `saveDock`-only — no re-fit dispatched — so it can't reform the #67 loop, and it seeds its baseline past the initial paint so the one-shot's `"server"` at load is untouched.

Verified: a real window resize → exactly one `"client"` emit that updates `_state` (grid `0 → 770` in the probe); initial paint still `"server"`; `test-state-source.R` now covers it (10 cases, 3 new). Full suite 141/0, no snapshot changes. Pushed; description updated.

nbenn:
Wiring the server-authoritative dock downstream (BristolMyersSquibb/blockr.dock#274) surfaced a concrete problem with this PR's `_state` model, and I think the fix belongs here rather than being worked around downstream.

**Focus rides the `_state` stream.** `callbacks.js` routes `onDidActivePanelChange` and `onDidActiveGroupChange` through `requestSync`, so a pure focus change re-emits the whole `_state` (the PR body lists "active panel / group change" as a `_state` gesture). `_state` already carries the active group in its `toJSON` payload, and the consumer reads focus straight out of it (`get_dock()[["activeGroup"]]`). So every tab click is a full-layout event on the heavy channel.

**Server-initiated focus churn leaks as `"client"`.** A `restore_dock` doesn't settle its active group synchronously — dockview activates each group in turn as it lays a multi-group layout out, and those `onDidActiveGroupChange` events fire in later tasks, after `withServerDriven`'s microtask clear. Each re-emits `_state` tagged `_state-source: "client"`, indistinguishable from a real tab-click, so the consumer folds every tick back into the board: the startup board-update burst in BristolMyersSquibb/blockr.dock#271 (~10-11 redundant commits per multi-view assembly). The 250 ms live-sync debounce downstream exists only to bound that burst — we're deleting it, and this coupling is the last blocker.

**Concrete ask: take focus off the `_state` trigger.** Drop `onDidActivePanelChange` / `onDidActiveGroupChange` from their `requestSync` calls, and keep them setting the dedicated `_active-panel` / `_active-group` inputs (they already do, right beside the `requestSync`). `_state` then re-emits only on *structural* gestures — move / add / remove / group add-remove / maximize / restore / sash / resize — and still snapshots the current active group in its payload at each of those. Focus becomes a payload field of `_state`, not an emission trigger.

**Why this rather than extending the provenance hold.** Tagging the async settle `"server"` (the `withServerDrivenThroughFlush` pattern from #78) keeps focus in the stream, but it has to outlast an async settle of unknown length — the very thing the fixed debounce was papering over. Decoupling sidesteps the timing question entirely: with no focus-triggered emit there is no client-leak to chase, sync or async. It also matches this PR's own thesis that `_state` is a discrete *structural gesture* stream — a tab click isn't an arrangement gesture, and it already has its own input.

**Downstream follow-up (ours, flagged for completeness).** With focus off the trigger, `_state`'s active group is a snapshot as of the last structural emit, so before serializing we'll force a fresh snapshot (a `_save-state` round-trip) or read `_active-group` directly. That's on the consumer side; noting it so the semantic shift — `_state` reflects structure, focus liveness rides the dedicated inputs — is explicit and deliberate.

nbenn:
Done — focus is off the `_state` trigger.

The `onDidActivePanelChange` / `onDidActiveGroupChange` handlers no longer call `requestSync`; they keep setting the dedicated `_active-panel` / `_active-group` inputs. Now `_state` re-emits only on structural gestures — move / add / remove / group add-remove / maximize / restore / sash / resize — and still snapshots the active group in its `toJSON` payload at each. Focus liveness rides the dedicated inputs.

That removes the leak: a server restore's async active-group settles — which fire in later tasks, after the provenance window closes — no longer re-emit `_state`, so there's nothing left to mis-tag `"client"`. The restore's structural layout still emits once, correctly `"server"` (via `onDidLayoutFromJSON`), so the startup board-update burst you cited loses its source and the downstream live-sync debounce can go.

Verified a focus change emits zero `_state` across repeated toggles and drives only `_active-panel` / `_active-group`. One nuance worth flagging: the first cross-group switch after load can produce a single `"client"` emit — but that is the `ResizeObserver` settling a genuine container-dimension change, not the focus path (synchronously the focus change emits zero), and it is one-time, not per-click. The `_state` active group is now a snapshot as of the last structural emit, exactly as you framed it — a consumer needing live focus reads `_active-group`, or forces a fresh snapshot before serialising.

The `test-state-source.R` suite gains a regression case (a same-group focus change → zero `_state`, `_active-panel` updated) and stays green — `restore` → `"server"`, panel close → `"client"` unchanged — full suite 142/0. Side effect: maximize now emits once rather than twice. NEWS and the PR description are updated.

nbenn:
Retract this request — it's the wrong fix, and `fae7693` should be reverted.

Taking focus off the `_state` trigger means a tab-click no longer refreshes `_state`, and serialization reads focus out of `_state` (via `view_data()` -> `get_dock()`), so focus is silently dropped on save: click a tab, save, reload, and you are back on the previous tab. Focus is meant to be persisted, so blanket-decoupling it from `_state` trades the startup burst for a real regression.

The burst is a provenance problem, not a "focus is in the stream" problem: a server restore's active-group settle leaks as `_state-source: "client"` because it fires after the `withServerDriven` window closes, so the consumer can't tell it from a genuine tab-click. The right fix is targeted — mute the focus `_state` emission only for the span of a server restore's settle, leaving it live for real user gestures — not remove focus from the stream wholesale. I'm instrumenting the startup emit sequence to scope that window precisely before proposing the concrete change, rather than hand over another blanket workaround.

nbenn:
Instrumented result: `fae7693` can be reverted outright — the decouple isn't needed, and it's a regression.

I traced the startup burst under a real browser with the consumer's `keep_foldable` and fold instrumented, debounce off, against the pre-decouple stack (this branch minus `fae7693` — i.e. provenance + init-tag + coalescing all present). Every startup focus emit is already `server`-tagged and dropped downstream:

- 2 views × 3 groups: 7 fold fires, **0 commits**, 0 client-tagged.
- 8 views × 2 groups: 17 fold fires, 128 `keep_foldable` decisions, **all `server` → DROP, 0 commits, 0 client** (including after a programmatic view switch).

So the provenance stack already absorbs the entire burst — `f8c5192` (provenance) + `77fbfff` (init render tagged server) + the discrete-gesture coalescing did the job. The `#271` burst was real on *pre-provenance* code and is since solved; the 250 ms consumer debounce is now vestigial. Taking focus off the `_state` trigger was fixing a burst that no longer exists, while introducing a real regression: a tab-click no longer refreshes `_state`, so serialization (which reads focus out of `_state` via `view_data()` -> `get_dock()`) silently drops it on save — click a tab, save, reload, you are back on the previous tab.

Concrete ask: **revert `fae7693`.** No replacement is needed — my earlier "mute focus during the restore" suggestion is moot, because the restore's active-group churn is already attributed to the server and dropped by the consumer, as the numbers above show. `_state` goes back to carrying focus live, and the consumer side is simply `#274` minus its (now dead) debounce: focus-sensitive fold, `keep_foldable` unchanged.

I'll confirm on CI — the faithful env, since my local fixtures render fast and a slow-block timing race is exactly what only CI would surface — and add an e2e for focus persistence (tab-click -> serialize -> active tab preserved), which doesn't exist today and is the gap that let this decouple pass green.

nbenn:
Reverted `fae7693` — branch is back to `82c6c04`.

You're right on both counts. Serialization reads focus out of `_state` (via `view_data()` → `get_dock()`), so decoupling silently dropped focus on save — a real regression — and your instrumented numbers show the startup burst is already fully absorbed by the provenance stack (`server`-tagged, dropped downstream, 0 client / 0 commits), so there was nothing left to fix. Focus is live in `_state` again; NEWS and the PR description are restored, and the regression test and focus paragraph are gone with the commit. No replacement, per your call.

The step I skipped was checking that a consumer reads focus *out of* `_state` before treating the emission as pure churn — which is exactly the gap the focus-persistence e2e you're adding (tab-click → serialize → active tab preserved) would have caught.

DivadNojnarg (commented):
Nice work overall: the resize-storm fix is real, the microtask coalescing argument for provenance is elegant, and the move/restore/resize test coverage is solid. I went through it with a fine comb (plus some headless verification against the bundled dockview dist). Almost everything below is one structural theme: the gesture allowlist now *defines* what "layout changed" means, so every dockview mutation outside the list silently stops reaching `_state`.

**Confirmed:**

- `srcjs/modules/callbacks.js:69-117`: `set_panel_title()` no longer updates `_state`. `setTitle` fires only `onDidTitleChange` (components.js just repaints the tab text) plus the now-disposed `onDidLayoutChange`; no hooked gesture fires, so after the proxy call `input$<id>_state` keeps the old title and a later save/restore reverts it. NEWS.md still lists `set_panel_title()` as a server `_state` producer, and the existing test only asserts the custom-message type, so nothing catches it.
- `srcjs/modules/callbacks.js:127`: floating-group move/resize never reaches `_state`. Shift-dragging a tab floats a group by default (`disableFloatingGroups` is never set), the overlay resizes via `.dv-resize-handle-*` rather than `.dv-sash`, its `onDidChangeEnd` is not subscribed, and the ResizeObserver only watches the main container. Floating geometry is silently lost on save/restore.
- `srcjs/modules/callbacks.js:154`: container-only resizes update `_state` but no longer re-fit widgets. `persist()` deliberately skips `refitWidgets()`, which is right for window resizes, but a bslib sidebar toggle or flex-parent change fires no window `resize`, so ECharts/plotly/DT panels keep stale pixel sizes. The old path dispatched a resize for every layout change.

**Plausible, worth a look or a test:**

- `srcjs/modules/callbacks.js:165-170`: initial `_state` for hidden/zero-sized containers (hidden tab, modal). The only `"server"`-tagged emit carries a zero-sized grid; the settled first layout lands in a later task and gets tagged `"client"`. The provenance test only covers a visible dock.
- `srcjs/modules/callbacks.js:126,143`: the container `pointerup` listener and the ResizeObserver are added per `renderValue` with no teardown (new in this PR; the old code only had `api.onDid*` subscriptions that die with the api). Static example apps never re-render, but a reactive `renderDockView` accumulates listeners holding stale `api`s that re-emit superseded state.
- `srcjs/modules/callbacks.js:150-154`: a container resize landing within the initial 150ms debounce window coalesces into the seeding pass and is recorded without `persist()`, so `_state` misses it until a second resize. Cheap fix: seed the baseline from the first-layout size instead of the first observation.
- Cross-PR note for the stack: `set_size` (#78) surfaces only through `onDidLayoutChange`, so once rebased onto this design it will resize the layout while `_state` goes stale; same for any future size-only command.

**Structural suggestion:** the title, floating and set_size gaps are one class. Since this PR already broke the feedback loop (re-fit no longer runs per-frame), a permanently-subscribed `onDidLayoutChange` debounced through the same `persist()` path the ResizeObserver uses would act as a safety net under the gesture allowlist: gestures give the crisp, provenance-correct boundaries; the debounced catch-all guarantees `_state` eventually converges for everything else. That closes the whole class instead of patching per op, at the cost of one trailing (correctly `"client"`-tagged) emit in rare cases.

Minor, non-blocking: the eight near-identical event hookups could collapse into a table; `flushScheduled`/`persisting` encode one state machine in two booleans with the try/finally bracket duplicated; the "removable at dockview >= 6.1.1" guard comment has no mechanism (version assertion or test) to ever retire it.

nbenn:
One thing worth folding in here: bump the `package.json` floor for `dockview-core` from `^4.10.0` to `^4.13.1`.

This PR updates `package-lock.json` but leaves the declared range at `^4.10.0`. That range was accurate while the `setSize` workaround was in place — dockViewR genuinely worked on 4.10 through 4.13. But once this PR sheds the workaround, the code depends on the upstream #1031 fix and no longer works correctly on 4.10.0–4.13.0 (the edge-drop rendering bug returns), so `^4.10.0` advertises a compatibility the code no longer has.

The lockfile pinning 4.13.1 covers the default install, but the range is the compatibility contract, not just a resolution hint: `^4.10.0` stops protecting the moment someone pins an older 4.x or 4.13.1 gets yanked — it would silently fall back to the buggy 4.13.0, whereas `^4.13.1` fails loudly, which is what you want.

For context: #83 (the 4.13.1 bump) deliberately kept `^4.10.0`, since with the workaround still present dockViewR really was compatible with 4.10+. This PR is where 4.13.1 turns from preferred into required, so it's the natural home for the one-line floor bump.

nbenn:
Worked through the review, @DivadNojnarg — here's what landed (the branch is rebased onto `main` now that the stack merged, all green):

- **Title**: the `_set-panel-title` handler now `saveDock`s inside `withServerDriven`, so `_state` carries the new title tagged `server`.
- **Container-only resizes**: the `ResizeObserver` re-fits widgets on every settle now, so a bslib sidebar toggle no longer strands ECharts/DT at stale sizes. Loop-safe — the observer watches the container, and a widget re-fit moves internal groups, not it.
- **Initial zero-sized grid**: replaced the `onDidLayoutChange` one-shot with the `ResizeObserver`'s first real-size observation, tagged `server`. The one-shot fired at width 0, never disposed (its `width > 0` guard failed), and a size-only settle never re-fires `onDidLayoutChange`, so it lingered and fired late (`client`) on the first gesture — which was also the stray emit landing on `set_panel_title`. `callbacks.js` now has no `onDidLayoutChange` subscription at all.
- **Listener teardown**: `setDockViewCallbacks` returns a cleanup; the factory disposes the old api and calls it before a re-render, so a reactive `renderDockView` stops accumulating stale-api pointerup / ResizeObserver listeners.
- **Resize inside the initial debounce window**: folds into the one-shot fix — the seed persists now, so a resize there lands as the settled size rather than being lost.

On the debounced-`onDidLayoutChange` safety net you floated: I went crisp-per-signal instead. Measuring showed `onDidLayoutChange` is a partial, unstable signal — it doesn't fire for add, maximize, or window-resize, and its per-gesture count isn't reproducible — so a shared catch-all wouldn't be reliable anyway. Each gap got its proper boundary.

**Open, and your call: floating groups.** Shift-drag-to-float is enabled by default, and a floated group's move/resize only surfaces through `overlay.onDidChangeEnd` → `onDidLayoutChange` — an internal object with no public creation hook — so its geometry is lost on save/restore. A hook against that overlay works today on our pinned 4.13.1, but it's untyped internal API, so it becomes a liability the next time we bump dockview off 4.x (which we'll want to eventually — the maximize/`toJSON` fix is in 6.1.1, and 7.x is already out). Unless floating groups are actually used downstream, my suggestion is to **disable them for now** (`disableFloatingGroups: true`): the feature comes off, the bug disappears, and there's nothing internal to maintain — we revisit if there's demand. Wire up the overlay hook instead, or disable?

nbenn:
Done — bumped the floor to `^4.13.1` (both `package.json` and the lockfile's declared range) in 87c9f10. Agreed this PR is its home: it's where the `setSize` workaround comes off, so 4.13.1 flips from preferred to required.

DivadNojnarg (commented):
Re-checked the branch after the rebase and the four follow-up commits, same harness as the original review (full test suite plus live headless verification against the state_source example).

**Verified fixed:**

- `set_panel_title` staleness: the handler persists inside its `withServerDriven`; live check shows `_state` carrying the new title immediately, tagged `"server"`.
- Re-render leaks: `setDockViewCallbacks` returns a teardown (pointerup listener, ResizeObserver, pending timer) and the factory runs it plus `api.dispose()` before rebuilding; gestures after a reactive re-render emit cleanly.
- Initial-tag race and the swallowed-first-resize, both closed by one change: the `firstLayout` one-shot is gone and the ResizeObserver's seeding branch persists the first real size server-tagged, so the tag no longer depends on landing in the render task and the first observation persists instead of silently seeding. Live check: initial `_state-source` is `"server"`.
- Container-only resizes now re-fit widgets (`refitWidgets()` on the debounced resize path, guarded).
- `dockview-core >= 4.13.1` is enforced at build time instead of by comment.

Full suite green, including the expanded `test-state-source.R`.

**Merge verdict: good to merge, with two follow-ups tracked:**

1. The floating-group/popout gap is the one remaining confirmed finding, and it is a regression vs main (the old catch-all persisted floating moves; now they never reach `_state`, and shift-drag-to-float is reachable by default). Nothing in the blockr stack uses floating groups today, so I would not block the merge on fixing it, but it should not live only in a PR comment: either an issue for a proper gesture hook (the floating overlay's change-end events, same pattern as the title handler), or `disableFloatingGroups: true` as the widget default until then.
2. Whichever of #78/#86 rebases onto this design needs the same one-liner the title handler got: `set_size` surfaces only through `onDidLayoutChange`, so post-merge it will resize the layout while `_state` goes stale; `saveDock` inside its `withServerDriven` closes it. I will leave a note on #78.

The remaining code-quality nits from the original review (event-hookup table, the two-boolean flush state) are cosmetic and fine to leave.

DivadNojnarg (approved):
See further comments for follow-up

PR #83, opened 2026-06-30
Bump bundled dockview-core to 4.13.1

### LINKED ISSUE #81: Bump dockview-core to 4.13.1 (carries the upstream #1031 fix)
We pin `dockview-core` 4.10.0 (2025-10-30), which predates mathuo/dockview#1031 — "Panel content does not render when dragged to extreme drop targets" under `defaultRenderer: 'always'`, which is the renderer this package uses. We work around it in `srcjs/modules/callbacks.js` by forcing `setSize` on the moved panel after every `onDidMovePanel`.

The fix shipped in dockview **v4.13.1** (2025-12-29) — v4.13.0 misses it by one commit. That's a minor bump still inside our existing `^4.10.0` range, so it's a lockfile update rather than a breaking change, and 4.13.1 is the last 4.x release (5/6/7.x are separate major-bump evaluations, out of scope here).

Bumping to 4.13.1 fixes the latent edge-drop rendering bug properly and unblocks dropping the `setSize` workaround — see the follow-up.

nbenn:
## Summary

- Bump the bundled `dockview-core` from 4.10.0 to 4.13.1 — the lockfile pin plus the rebuilt webpack bundle. 4.13.1 satisfies the existing `^4.10.0` range, so `package.json` is unchanged; the JS-side compatibility is unchanged across all of 4.x, so this is a lockfile + artifact bump, not an API break.
- 4.13.1 carries the upstream fix for panel content failing to render when a panel is dragged to an extreme drop target (mathuo/dockview#1031) — the bug the `setSize` workaround in `srcjs/modules/callbacks.js` currently papers over.
- The workaround stays in this PR; dropping it (and coalescing `_state` on a microtask) is the follow-up #82.

Stacked on #79, whose app-test grid-geometry removal keeps this bump from churning the snapshots. Retarget to `main` once #79 lands.

Fixes #81


DivadNojnarg (approved):
@nbenn LGTM thanks

PR #86, opened 2026-07-09
Align move_panel placement with add_panel vocabulary

### LINKED ISSUE #85: Expose movePanel as a proxy command
The proxy command family covers `add_panel`, `remove_panel`, `select_panel` and `set_panel_title` (plus `set_size` on the `76-set-size` branch), but there is no way to move an existing panel — dockview-core's `movePanel` API is not exposed.

The consumer case is programmatic rearrangement in blockr.dock: server-side callers (an LLM assistant among them) express layout intents as atomic panel ops with placement hints, and "move panel X next to panel Y" is the one op in that algebra with no dockViewR counterpart. The interim workaround — `remove_panel` followed by `add_panel` with a `position` — re-creates the panel rather than moving it, losing the panel instance and flickering.

Would a `move_panel(proxy, id, position)` be worth adding, with `position` speaking the same vocabulary as `add_panel` (`referencePanel` / `referenceGroup` plus `direction`)? Happy to put together a PR along the lines of the existing proxy commands if that shape sounds right.

nbenn:
## Summary

- `move_panel()` now takes a `position` list — a `referencePanel` or `referenceGroup` plus a `direction` (`above` / `below` / `left` / `right` / `within`) and an optional `index` — the same placement vocabulary `add_panel()` and `panel()` already use, replacing the old scalar `position` / `group` / `index` arguments.
- On the JS side the reference resolves to the target group and `direction` maps onto dockview's `panel.api.moveTo()` `Position` (`within` → `center`). The panel is moved, not recreated, so its instance survives — unlike the `remove_panel()` + `add_panel()` workaround the issue describes, which drops the instance and flickers.
- A server-side caller that expresses layout intents as placement hints (an assistant driving blockr.dock, for one) can now use one grammar for both adding and moving panels.

## Migration

| before | after |
| --- | --- |
| `move_panel(dock, id, group = "b", position = "right")` | `move_panel(dock, id, position = list(referencePanel = "b", direction = "right"))` |
| `move_panel(dock, id, group = "b", position = "center")` | `move_panel(dock, id, position = list(referencePanel = "b", direction = "within"))` |

The `direction` vocabulary matches `add_panel()`, so it differs from the old scalar `position`: `top` / `bottom` become `above` / `below`, and `center` becomes `within`. The reference panel/group moves from the `group` argument into `position$referencePanel` (or `position$referenceGroup`), and `index` moves into the list.

Stacked on #78 — base is `76-set-size`; retarget to `main` once that lands.

Fixes #85


DivadNojnarg (commented):
Reviewed with the same fine comb as #80, including checking the bundled dockview `moveTo` on both old and new bundles. The design is right: the R validation is genuinely shared (`process_panel_position()` serves `panel()`, `add_panel()` and now `move_panel()`), every in-repo caller was migrated, old-style `group=`/`index=` calls fail loudly since there is no `...` to swallow them, and the `directionToPosition` map covers the R vocabulary exactly. blockr.dock is unaffected today (`op_move_panel` still does remove+add, waiting on #85, and `hint_to_position()` already emits the new shape, so adoption will be nearly drop-in).

**Findings, most severe first:**

- `R/utils.R:130-138`: the two most likely misuse cases produce cryptic base-R errors instead of validation messages (verified by execution). Omitting `direction` (`move_panel(dock, id, position = list(referencePanel = "a"))`) dies with `argument is of length zero` from `if (!(NULL %in% valid_directions))`, and the old-API migration call `position = "right"` dies with `subscript out of bounds`. One `is.list(position)` guard plus a `direction`-presence check with real messages fixes both; since the helper is shared, `add_panel()` inherits the fix.
- `tests/testthat/test-proxy.R` / `inst/examples/move_panel/app.R:106`: same-group tab reorder (`direction = "within"` + `index`) is covered by no test or example anymore. The example's first button changed from the old index-reorder to a split-out, and the vignette documents `index` as "used with direction = within" without anything pinning that dockview honors it. A regression in the `center`+`index` path would go undetected.
- `srcjs/modules/proxy.js:128`: `index` is silently ignored for directional moves (`left/right/above/below` create a new group where index is meaningless). Cheap to document, or to warn R-side when `index` is combined with a non-`within` direction.
- `DESCRIPTION:3`: no version bump for a documented breaking change (0.3.0 on base and head). blockr.dock pins `dockViewR (>= 0.2.1)`, so nothing guards the break at install/check time; the stack's merge to main should bump minor.
- `srcjs/modules/proxy.js:115-123`: when both `referencePanel` and `referenceGroup` are supplied, `referenceGroup` silently wins; neither side rejects or documents the ambiguity.
- Doc gap, not a bug: a reference-free `position = list(direction = "right")` is accepted and, per the bundled `moveTo` (`group ?? addGroup({direction})`), moves the panel into a new group at the dock edge in that direction. Coherent (mirrors `add_panel` without a reference) but undocumented; worth a sentence in `?move_panel` so it reads as intended.

**Checked and cleared, to preempt concern:** the apparent loss of the old index-only same-group reorder is not a regression. `move_panel(dock, id, index = 3)` called `moveTo({index})` with no group, and in both the old and new bundles that code path is identical: it creates a new right-edge group rather than reordering, so the old documented "moved to the latest index" behavior never matched dockview. The new grammar can express same-group reorder correctly (`referencePanel` = a sibling, `direction = "within"`, `index`), which is another reason to add that missing test. Also verified `getGroup()` and `panel.api.group` both hand `moveTo` the same DockviewGroupPanel type, so the `referenceGroup` branch is sound.

nbenn:
Thanks for the thorough pass — every finding was actionable. Addressed in the push just now (rebased onto `main` after #78 landed):

- **`R/utils.R` cryptic errors** — `process_panel_position()` now guards `is.list(position)` (so `position = "right"` reports "`position` must be a list") and treats a missing `direction` as a validation failure rather than letting `NULL %in% valid_directions` blow up. Both messages flow through the shared helper, so `add_panel()` inherits them.
- **`within` + `index` coverage** — the first example button is now a genuine same-group reorder (`referencePanel = "2", direction = "within", index = 2`), and its app-test snapshot captures the reordered `views`, so a regression in the `center` + `index` path fails the test. Added a unit case exercising `referenceGroup` + `within` + `index` too.
- **`index` ignored for directional moves** — `process_panel_position()` now warns when `index` is combined with a non-`within` direction, and `?move_panel` spells out that `index` only applies to `within`.
- **Version bump** — `DESCRIPTION` is now `0.4.0`, so downstream pins can guard the break.
- **`referencePanel` + `referenceGroup` ambiguity** — supplying both is now an error rather than a silent `referenceGroup` win.
- **Reference-free direction** — documented in `?move_panel`: with no reference the panel moves to a new group at the dock edge in `direction`, mirroring `add_panel()`.

On the cleared item: agreed the old `index`-only path never truly reordered, so removing it loses nothing — and the new grammar's same-group reorder is what the first example/test now pins.

PR #90, opened 2026-07-19
Give a restore its own _state-source value

### LINKED ISSUE #89: Distinguish a restore from other proxy ops in `_state-source`
The `_state-source` input (added in #70) tags an echo `"server"` when a proxy op provoked it and `"client"` for a user gesture. The flag is raised by the shared `C()` wrapper around every handler — `add-panel`, `move-panel`, `set-size`, `select-panel`, `move-group`, `restore-state` — so a consumer cannot tell a restore's echo apart from the echo a programmatic move or resize produces.

That distinction is exactly what a downstream consumer needs. blockr.dock's grid mirror commits the client's settled layout back to the board, and it needs opposite behaviour for the two cases:

- A `restore-state` echo is a redundant replay of the layout the server just pushed, so it must be **skipped** — committing it re-persists what we already hold, and any in-flight frame is a transient nobody authored.
- A `move-panel` / `set-size` / `add-panel` echo carries the geometry the client actually realised, which is the *only* server-side record of where the panel landed or how it was sized, so it must be **committed**.

With only `"server"` vs `"client"`, neither a blanket skip nor a blanket commit is correct. A blanket skip drops the move / resize / add geometry, so it never persists. A blanket commit re-persists the restore replay — and because a `select-panel` fired after a restore echoes the client's rendered sizes, it also drifts the stored grid off the authored value on a plain load, so a save→reload no longer reproduces the saved geometry.

Proposal: give `restore-state` its own source value, `_state-source == "restore"`, leaving `"server"` for the other proxy ops and `"client"` for user gestures. A consumer then skips only `"restore"` echoes and commits the rest. (A separate boolean input would also work; a distinct `_state-source` value keeps it to one signal.)

blockr.dock is currently working around this with a one-shot restore latch on its own side, which cannot scope the full restore realisation (hence the `select-panel` drift above). The finer tag would let it drop the workaround and skip the whole restore burst cleanly.

nbenn:
## Summary

- `input$<dock_id>_state-source` now reads `"restore"` for the `_state` a `restore_dock()` echoes back, distinct from the `"server"` tag on every other proxy op (add / remove / move / resize / set title) and `"client"` on a user gesture.
- Why the two-value tag was not enough: a restore only replays the layout the server just pushed, whereas a move / resize / add echo is the *only* server-side record of the geometry the client actually realised. With both tagged `"server"`, a consumer could neither skip the redundant restore replay (a blanket commit re-persists it and drifts the stored grid off the authored value, so a save→reload no longer reproduces the saved geometry) nor keep the realised geometry (a blanket skip drops it). The distinct `"restore"` value lets a consumer skip only the replay.
- Mechanism: the per-id provenance flag now holds a source label rather than a boolean — `withServerDriven` sets `"server"`, a new `withRestore` sets `"restore"`, and the `_restore-state` handler uses `withRestore`. The causal-microtask attribution (no wall-clock windows) is unchanged.
- The compiled `inst/htmlwidgets/dockview.js` is rebuilt against the locked `dockview-core` 4.13.1. `main`'s committed bundle had regressed to a 4.10.0 build (rebuilt against a stale `node_modules`) while `package.json`, `package-lock.json` and NEWS all pin 4.13.1 — so the shipped widget lacked the upstream mathuo/dockview#1031 fix (content not rendering on a drag to an extreme drop target) that the removed `setSize` workaround relied on. The second commit isolates that rebuild from the provenance change.
- Downstream, BristolMyersSquibb/blockr.dock#343 tracks dropping the one-shot restore-latch workaround this replaces.

Fixes #89


nbenn:
Closing — the `_state-source` provenance this PR extends is being retired rather than extended. blockr.dock (the only consumer) no longer reads it (BristolMyersSquibb/blockr.dock#344, CI-green with the tag removed), and dockview-core v7 ships the same idea natively (`onDidMutateLayout` with user-vs-api origin). See #92 to retire the layer and #91 to emit the restore's `_state` on `onDidLayoutFromJSON` instead.

Leaving the `89-restore-source` branch in place for now — blockr.dock#344 still pins it until its own cleanup lands.

Issue #92, opened 2026-07-20
Retire the _state-source provenance layer


nbenn:
The `_state-source` input (added in #70 / #71, with the never-merged `"restore"` value proposed in #89 / #90) exists so a consumer can tell server-driven echoes from user gestures. It should be retired rather than extended, for two reasons:

- **The sole consumer no longer uses it.** blockr.dock's grid mirror stopped reading `_state-source`; it keeps restored geometry correct with a settle-focus plus an empty-grid guard instead (BristolMyersSquibb/blockr.dock#344, CI-green with the tag removed).
- **Upstream now ships the same idea natively.** dockview-core v7 added layout-mutation transaction events (`onWillMutateLayout` / `onDidMutateLayout`) that carry user-vs-api origin (mathuo/dockview#1317, #1328, #1335). On dockview-core v7 the hand-rolled `_state-source` / `withServerDriven` / `withRestore` machinery in `proxy.js`, `handlers.js` and `callbacks.js` duplicates it.

Remove `_state-source` and its wrappers. #91 covers the one thing a consumer still wants from a restore — its settled `_state`, keyed on `onDidLayoutFromJSON`; everything else the tag did is either unused or now native.

PR #93, opened 2026-07-20
Emit only a settled _state on restore and re-init

### LINKED ISSUE #91: Restore emission hygiene: surface a restore as a single settled _state, never a transient
## What we want

When the server restores a layout (`_restore-state` → `api.fromJSON`), the `_state` input a consumer sees must only ever hold a **valid, settled layout**. It must never surface an intermediate frame — an empty layout while the dock is (re)initialising, or a structure caught mid-build. A consumer that mirrors `_state` into persistent storage can then treat every `_state` it receives as real geometry, with no per-frame guard and no provenance tag.

## Why — the concrete failure

blockr.dock mirrors `_state` into a board's stored grid. On an import/reload round-trip the restored board came back with an **empty grid**: an empty `_state` reached the mirror during the re-init and got committed as the layout. It is currently patched consumer-side with an empty-frame guard plus a forced settle (BristolMyersSquibb/blockr.dock#344). That guard only exists because dockViewR surfaces a non-settled `_state`; the clean fix is to not surface it.

## What is already correct — so the fix is narrow

`fromJSON()` in dockview-core is synchronous: it `clear()`s, builds every group and panel, then fires `onDidLayoutFromJSON` at the very end (`dockviewComponent.js`). The layout-change events it fires along the way are coalesced through an `AsapEvent` (`baseComponentGridview.js`, `events.js`), so the *synchronous* restore already collapses to one settled emission via `requestSync` → `saveDock`. The job is **not** "coalesce the restore" — that is done. The job is the frame(s) that escape *around* it.

## The fix

Gate `_state` emission on the restore boundary, keyed on `onDidLayoutFromJSON` (the "structure fully built" signal):

1. In `handlers.js`, bracket the `_restore-state` handler: raise a `restoring` flag before `api.fromJSON(state)`, and clear it in a one-shot `onDidLayoutFromJSON` subscription that performs the single `saveDock`.
2. In `callbacks.js`, make `requestSync` a no-op while `restoring` is set (alongside the existing `flushScheduled` / `persisting` guards), so nothing emits between the start of the restore and `onDidLayoutFromJSON`.
3. Cover the empty re-init frame explicitly: **trace where the pre-restore empty `_state` actually originates** — the initial empty render of the widget vs. `fromJSON`'s own `clear()` — and make sure it falls inside the gated window (or is otherwise not emitted). This is the frame #344 observed; do not assume its source, confirm it.

Leave the post-restore async tail alone: the ResizeObserver "seeding" `saveDock` (first real container size) and the `requestAnimationFrame` position pass fire *after* `onDidLayoutFromJSON` and carry the real settled geometry — they are legitimate emissions, not transients.

## Explicitly not this

Do not add a `_state-source` / provenance value for the consumer to filter on. That was tried and retired (#89 / #90 closed; #92 removes the remaining machinery). The whole point is that the consumer needs no filtering because a transient never reaches it. (dockview-core v7's native `onDidMutateLayout` origin tagging is a separate matter; this issue is only restore emission hygiene.)

## Done when

- A `fromJSON` restore surfaces exactly the settled layout on `_state`; no empty or partial `_state` is emitted during a restore or re-init.
- Asserted against the `_state` JSON export that gates CI: after a restore the exported layout equals the input, and an empty/partial layout is never exported mid-restore.
- Existing app-tests stay green.

nbenn:
## Summary

- `input$<dock_id>_state` could surface a transient during a restore or a widget re-render. The eager initial `saveDock` emitted an empty grid, and the pre-geometry render flush emitted a zero-sized structure; a consumer mirroring `_state` into storage committed those, so a reload round-trip could land an empty or partial layout (patched consumer-side today with an empty-frame guard).
- `_state` emission is now gated on the settle boundary, so it only ever surfaces a valid, settled layout — no per-frame guard or provenance tag needed:
  - A restore raises a `restoring` flag so `requestSync` is a no-op across `fromJSON`'s teardown/rebuild, then emits the single settled layout from a one-shot `onDidLayoutFromJSON`.
  - The initial render drops the eager empty `saveDock` and holds `_state` until the ResizeObserver reports real geometry — which now also binds the panels' inputs.
- Because `_state` is now absent until that first settle, the `_state`-derived getters (`grid_shape()`, `get_active_views()`, `get_active_panel()`) return `NULL` for a not-yet-ready dock instead of erroring; the `demo` app-test waits for the input before snapshotting the grid.
- Regression test drives the `state_source` example through a re-render and a restore and asserts every `_state` emission is a settled frame (panels + real geometry). It fails on the pre-fix bundle (the empty and zero-sized frames) and passes on the fix.
- Bundle rebuilt against the locked `dockview-core` 4.13.1; the committed bundle had drifted to a stale 4.10.0 build, so the bundle diff is the version bump plus this change.

Supersedes the earlier `_restored-state` signal approach: per the rewritten issue, the consumer needs no separate channel or filtering because a transient never reaches it. Pairs with #92 (retire the `_state-source` provenance layer).

Fixes #91


srcjs/modules/handlers.js
```diff
@@ -58,9 +56,19 @@ const setShinyHandlers = (id, mode, api) => {
...
+  // requestSync stays a no-op while fromJSON tears down and rebuilds, then emit
+  // the single settled layout from a one-shot onDidLayoutFromJSON (fired
+  // synchronously at the end of fromJSON), server-tagged. No empty or
+  // intermediate frame escapes.
   Shiny.addCustomMessageHandler(id + '_restore-state', (m) => {
-    withServerDriven(id, () => restoreDock(id, m, api));
+    setRestoring(id, true);
+    const settled = api.onDidLayoutFromJSON(() => {
+      settled.dispose();
+      withServerDriven(id, () => saveDock(id, api));
```
DivadNojnarg: **Restore no longer rebinds panel inputs/outputs.**

`restoreDock()` still calls `Shiny.unbindAll()` and `fromJSON` re-creates the panel bodies via `innerHTML`, but the settled emit here calls `saveDock()` directly instead of going through `persistState()` -- which is the only place `Shiny.initializeInputs` / `Shiny.bindAll` runs (`callbacks.js:19-20`). `refitWidgets()` is no longer dispatched after a restore either.

Repro with `inst/examples/serialise`: Save, then Restore. `sliderInput("obs")` and `plotOutput("distPlot")` come back as dead DOM -- the slider no longer reports to the server and the plot never re-renders.

The emit needs the full persist path (or an explicit bind + refit) rather than the bare `saveDock`.

srcjs/modules/handlers.js
```diff
@@ -58,9 +56,19 @@ const setShinyHandlers = (id, mode, api) => {
...
 
-  // Restore layout
+  // Restore layout. Gate `_state` across the whole rebuild: raise `restoring` so
+  // requestSync stays a no-op while fromJSON tears down and rebuilds, then emit
+  // the single settled layout from a one-shot onDidLayoutFromJSON (fired
+  // synchronously at the end of fromJSON), server-tagged. No empty or
+  // intermediate frame escapes.
   Shiny.addCustomMessageHandler(id + '_restore-state', (m) => {
-    withServerDriven(id, () => restoreDock(id, m, api));
+    setRestoring(id, true);
```
DivadNojnarg, line 65: **`restoring` is never cleared when `fromJSON` throws.**

In dockview-core 4.13.1, `fromJSON` reverts and rethrows *before* `_onDidLayoutFromJSON.fire()`. So on any bad state -- `restore_dock(proxy, list())`, a layout serialised from a different panel set -- the one-shot listener never runs: `restoring[id]` stays `true` forever, every subsequent `requestSync` no-ops, and `_state` never updates again for that dock. The listener leaks too.

Wants a `try`/`finally` around `restoreDock(id, m, api)` that disposes `settled` and calls `setRestoring(id, false)` on the throw path.

srcjs/modules/callbacks.js
```diff
@@ -154,12 +164,8 @@ const setDockViewCallbacks = (id, api) => {
...
         try {
           if (seeding) {
-            // First real size. The synchronous render captures the layout while
-            // the grid is still zero-sized, and a size-only settle never re-fires
-            // onDidLayoutChange, so this is the only reliable signal that the
-            // initial layout landed. Server-tagged -- the render is
-            // server-initiated -- completing the zero-sized grid.
-            withServerDriven(id, () => saveDock(id, api));
+            seeded = true;
+            withServerDriven(id, () => persistState());
```
DivadNojnarg, line 204: **Panel binding is now gated on non-zero container geometry.**

Binding moved into this ResizeObserver seeding branch, but the observer returns early while the container measures 0x0. A dock first rendered inside a hidden `navset_hidden` tab or a false `conditionalPanel` therefore never binds its panels' inputs/outputs -- pre-PR, `onDidAddPanel -> requestSync -> persistState` bound them regardless of visibility.

Gating the `_state` *emit* on real geometry is the point of the PR and is fine; coupling *binding* to it is the part that regresses. Worth keeping binding on the panel-add path and letting only the emit wait for `seeded`.

srcjs/modules/handlers.js
```diff
@@ -29,9 +29,7 @@ const restoreDock = (id, state, api) => {
       }
     });
   }
-  let res = api.fromJSON(state);
-  Shiny.setInputValue(id + '_restored', true, { priority: 'event' });
-  return res;
+  return api.fromJSON(state);
```
DivadNojnarg, line 32: **Public `input$<id>_restored` removed without notice.**

This input is documented in `NEWS.md` for 0.3.0. It is deleted here with no replacement and no Breaking-changes entry, so downstream `observeEvent(input[["dock_restored"]], ...)` silently stops firing rather than erroring.

Either keep it (it is one `setInputValue` in the settled callback, where it would now be strictly more correct than before) or add a Breaking-changes NEWS bullet pointing consumers at `_state`.

srcjs/modules/handlers.js
```diff
@@ -58,9 +56,19 @@ const setShinyHandlers = (id, mode, api) => {
     saveDock(id, api)
```
DivadNojnarg, line 56: **Direct `saveDock` handlers bypass the new `seeded` gate.** (lower confidence)

`_save-state` here, plus `_set-size` and `_set-panel-title`, write `_state` unconditionally without consulting `seeded`. An early `save_dock()` or `set_panel_title()` -- e.g. from an `observe()` that fires before the ResizeObserver reports real geometry -- can still publish a pre-geometry frame, which contradicts the NEWS promise that `_state` only ever surfaces a settled layout.

Narrower than the paths the PR fixes, but it is the same class of transient.

R/dock_state.R
```diff
@@ -167,8 +180,11 @@ get_active_views <- function(dock) {
...
   active_group <- get_active_group(dock)
   active_views <- get_active_views(dock)
-  # We need no check since there is always an active group
-  # and an active view.
+
+  if (is.null(active_group) || is.null(active_views)) {
+    return(NULL)
+  }
+
   active_views[[active_group]]
```
DivadNojnarg, line 188: **`get_active_panel()` can still error.** (lower confidence)

The new guards cover the `NULL` case only. If the active group has no active view while other groups do, `active_views` is a non-`NULL` list that lacks the `active_group` key, and this line raises `subscript out of bounds` -- which is exactly the invariant the deleted "We need no check since there is always an active group and an active view" comment was asserting.

`active_views[[active_group]]` guarded with a `%in% names(active_views)` check (or `active_views[[active_group]] %||% NULL` semantics) would close it.

DivadNojnarg (commented):
@nbenn Reviewed the diff plus the regenerated bundle. The core idea (gate emission on the settle boundary) looks right, and the bundle is genuinely in sync with `srcjs`. Six findings inline, two of which I'd treat as blockers:

- Restore no longer rebinds panel inputs/outputs, so `inst/examples/serialise` has a dead slider/plot after Save -> Restore.
- The `restoring` flag is never cleared when `fromJSON` throws, which wedges `_state` permanently for that dock.

The rest are a visibility-gating regression on binding, an undocumented removal of the public `input$<id>_restored`, direct `saveDock` handlers bypassing the new `seeded` gate, and a residual error path in `get_active_panel()`.

Verified as fine: bundle matches `srcjs` (rebuilt, `_restored` gone), the restore path keeps `_state-source` as `"server"`, the maximized-group re-entrancy guard still holds during a restore (`isRestoring` cleared only after `saveDock`), and the other `_state` readers (`get_panels_ids`, `get_groups_ids`, `get_groups_panels`) already degrade to `NULL`/`list()` on an absent state.

srcjs/modules/handlers.js
nbenn: Confirmed and fixed in 418d7fe — good catch, this was a real regression.

Reproduced it on the pre-fix build exactly as you described: bring `test` to the front (it starts behind panel 2, so its content is not in the DOM until activated), Save, Restore — the slider is still rendered but `$("#obs").data("shiny-input-binding")` is gone, and `set_inputs(obs = 781)` leaves the server on the stale 500.

The settled emit now runs the full `persistState()` (so the rebind happens) plus `refitWidgets()`. One thing that fell out of fixing it: doing that synchronously inside `onDidLayoutFromJSON` still does not bind, because the rebuilt panel bodies are mid-attach while the event fires. It has to run on the microtask after it — the same point the coalesced gesture flush used to land on. Regression test added (`a restore rebinds the panels' inputs and outputs`), verified failing on the pre-fix bundle at 500 and passing at 781.

srcjs/modules/handlers.js
nbenn: Confirmed and fixed in 418d7fe.

Verified the ordering in the bundled 4.13.1 source: the `catch (err)` block ends `this.clear(); throw err;` and `_onDidLayoutFromJSON.fire()` is only reached on the success path, so the settle callback genuinely never runs for a corrupt layout.

The handler now clears on the throw path only (`catch` + rethrow) — the success path is left to the settle callback, so the gate stays shut until the settled layout is actually published rather than being lowered early by a `finally`. The listener is no longer per-call: it moved to `setDockViewCallbacks`, where `persistState` lives, so it dies with `api.dispose()` and there is nothing to leak.

Regression test added (`a failed restore does not wedge _state`): `restore_dock(proxy, list())`, then an unplaced `add_panel()` — note the failed restore leaves the dock cleared, so the follow-up gesture cannot reference a pre-restore panel. Fails pre-fix (the new panel never reaches `_state`), passes now.

srcjs/modules/callbacks.js
nbenn: Agreed, and fixed in 418d7fe — the coupling was mine and it was wrong.

Binding is out of the gate entirely now: `requestSync` no longer consults `seeded`, and `persistState` binds unconditionally. Only the emit waits, and it waits inside `saveDock` rather than at the call site. So a dock that never gets geometry still binds its panels on the panel-add path, exactly as before this PR.

srcjs/modules/handlers.js
nbenn: You are right, that was an oversight rather than a decision — restored in 418d7fe, no breaking change.

It now fires from the settle callback, after the settled layout has been published and the panels rebound, instead of immediately after `fromJSON()` returned. That makes it strictly more useful than before: an `observeEvent(input[["dock_restored"]], ...)` can read the restored layout out of `_state` directly rather than racing it. NEWS gets a bug-fix bullet for the timing change.

srcjs/modules/handlers.js
nbenn: Fixed in 418d7fe — agreed it is the same class, and closing it turned out to simplify things.

The gate moved out of `requestSync` and into `saveDock`, which now early-returns unless the dock is seeded and not mid-restore. That makes it the single choke point, so `_save-state`, `_set-size` and `_set-panel-title` are covered without each having to opt in. `seeded` moved next to `restoring` as a per-id flag for that, and `setDockViewCallbacks` resets it per render so a re-render does not inherit the previous dock's gate.

An early `save_dock()` is therefore dropped rather than publishing a pre-geometry frame; the ResizeObserver seed emits the real layout immediately after, so the consumer still converges — and it now matches what NEWS promises.

R/dock_state.R
nbenn: Fixed in 418d7fe.

Collapsed the check to membership, which covers the missing-key case you describe as well as the two `NULL` ones:

```r
if (!isTRUE(active_group %in% names(active_views))) {
  return(NULL)
}
```

nbenn:
Thanks — all six are fixed in 418d7fe, replies inline on each thread.

Both blockers were real and are now covered by regression tests that I checked fail on the previous bundle: the restored slider stayed at 500 instead of 781, and after a corrupt restore the newly added panel never reached `_state`.

The one structural change worth flagging beyond the individual fixes: gating moved out of `requestSync` and into `saveDock` itself, which early-returns unless the dock is seeded and not mid-restore. That was the cheapest way to close your fifth point, and it also decouples binding from the geometry gate (your third), since `persistState` can now bind unconditionally and let the emit gate itself. `seeded` became a per-id flag next to `restoring`, reset per render so a re-render does not inherit the previous dock's gate.

One nuance that cost me a round: running the settled persist synchronously inside `onDidLayoutFromJSON` still does not rebind, because the rebuilt panel bodies are mid-attach while the event fires. It runs on the microtask after it instead.

nbenn:
@DivadNojnarg good to go or do you want to have another look?

srcjs/modules/callbacks.js
```diff
@@ -59,7 +76,32 @@ const setDockViewCallbacks = (id, api) => {
...
+  // still being attached, so the persist waits for the microtask after it -- the
+  // same point the coalesced gesture flush would have run.
+  api.onDidLayoutFromJSON(() => {
+    queueMicrotask(() => {
+      setRestoring(id, false);
+
+      persisting = true;
+      try {
+        withServerDriven(id, () => persistState());
+        refitWidgets();
```
DivadNojnarg: Optional: this `refitWidgets()` is not behind the `isSeeded(id)` guard you added to the gesture flush in `12e183bd` (line 70), and your rationale there applies here too — dispatching a `resize` into a zero-geometry dock restarts the ResizeObserver's debounce and delays the very observation that seeds the layout.

Only reachable when a restore lands before first geometry (`restore_dock()` at startup, or a dock in a hidden tab), and there is nothing to re-fit in that case anyway, so it is the same `if (isSeeded(id))` for consistency rather than a bug you can hit in the common path.

srcjs/modules/proxy.js
```diff
@@ -323,10 +323,36 @@ const runServerDriven = (id, value, fn) => {
...
+};
+
+const isRestoring = (id) => restoring[id] === true;
+
+const setRestoring = (id, value) => {
+  restoring[id] = value;
+};
+
 const saveDock = (id, api) => {
+  if (!isSeeded(id) || isRestoring(id)) return;
```
DivadNojnarg, line 351: Optional, doc-only: this makes `save_dock()` a silent no-op on a dock that has never been seeded, where it previously published a (zero-sized but structurally valid) layout.

It is the right tradeoff — it converges as soon as the dock is shown, and a pre-geometry frame is exactly what this PR is removing — but it is a user-visible behaviour change for anyone calling `save_dock()` against a dock in a not-yet-opened tab, and the R getters returning `NULL` there is the same story. Worth a sentence in the NEWS entry so it reads as intended rather than as a missing emit.

DivadNojnarg (approved):
@nbenn Had another look at `418d7fe` + `12e183bd`. All six are genuinely fixed, and the two places you pushed back you were right to. Good to go.

What I checked rather than took on trust:

- **Rebind** — the settle callback now runs the full `persistState()` plus `refitWidgets()` on the microtask after `onDidLayoutFromJSON`. Your point about the rebuilt panel bodies still being mid-attach when the event fires synchronously matches what I saw in the 4.13.1 source, so the microtask hop is load-bearing, not defensive.
- **Throw path** — `try`/`catch` + rethrow in the handler. Declining the `finally` I suggested is the better call: a `finally` runs after `fromJSON` returns but *before* the settle microtask, so it would reopen the gate early and let a stray gesture publish ahead of the settled layout. Moving the listener into `setDockViewCallbacks` also kills the per-call leak.
- **Binding vs geometry** — `requestSync` no longer consults `seeded` and `persistState` binds unconditionally, so a dock in a hidden tab keeps working inputs.
- **Gate in `saveDock`** — better than what I proposed. One choke point covers `_save-state` / `_set-size` / `_set-panel-title` without each opting in, and the per-render `setSeeded(id, false)` stops a re-render inheriting the previous gate.
- **`_restored`** — back, and firing after the settled publish is strictly more useful than the old timing. NEWS bullet is right.
- **`get_active_panel()`** — the membership collapse covers all three cases; `NULL %in% names(x)` gives `logical(0)`, which `isTRUE()` folds to `FALSE`, so the `NULL` group and `NULL` views paths still return `NULL`.

Both new tests are real rather than shape-only: the rebind test drives an actual input round-trip (781 vs the stale 500), and the failed-restore test asserts through `panel_ids`, which derives from `_state` — a wedged gate would fail it. Bundle carries all three new pieces.

Two residuals inline, both optional and neither worth holding the merge for.

srcjs/modules/callbacks.js
nbenn: Agreed — inconsistent of me to guard one and not the other. Same `if (isSeeded(id))` applied in d2232f3.

You are right that it is unreachable in the common path, but it is reachable exactly where it would hurt most: a `restore_dock()` at startup is the case where delaying the seeding observation also delays the first `_state`, which is the thing this PR exists to get right.

srcjs/modules/proxy.js
nbenn: Fair — that deserved to be stated rather than left to be inferred. Added to the NEWS entry in d2232f3, covering both halves: `save_dock()` publishing nothing and the `_state`-derived readers returning `NULL` until the dock is first shown, where a hidden-tab dock used to report a zero-sized layout.

Framed as "not laid out yet" rather than a missing emit, with the note that it converges as soon as the dock becomes visible.

PR #94, opened 2026-07-21
Retire the _state-source provenance layer

### LINKED ISSUE #92: Retire the _state-source provenance layer

nbenn:
## Summary

- Removes `input[["<dock_id>_state-source"]]` and the `withServerDriven` provenance machinery in `proxy.js` / `handlers.js` / `callbacks.js` behind it. The tag distinguished a server-driven `_state` echo from a user gesture so a consumer could ignore a layout it had just pushed; no consumer relies on it any more (blockr.dock keeps restored geometry correct with a settle-focus plus an empty-grid guard, BristolMyersSquibb/blockr.dock#344), and a future dockview-core will report the same programmatic-vs-user origin natively through its layout-mutation transaction events (mathuo/dockview#1317, #1328, #1335), so the hand-rolled layer is not worth carrying until then.
- The removal is behaviourally inert beyond the retired input: `withServerDriven` only raised and lowered a flag that `saveDock` read to pick the tag, and forwarded a return value nobody used. Unwrapping the handlers and the two `persistState()` call sites changes no ordering and no emit count. The `seeded` / `restoring` / `persisting` gates from #93 — including `saveDock`'s self-gating and the `onDidLayoutFromJSON` settle boundary — are untouched.
- Renames the `state_source` example to `state` and `test-state-source.R` to `test-state-hygiene.R`, since both now exercise the `_state` input rather than the retired tag; drops the provenance-only assertions and keeps the `_state` emission-hygiene coverage.
- Rebuilds the bundle against the locked dockview-core 4.13.1.

### Migration

| Before | After |
| --- | --- |
| Read `input[["<id>_state-source"]]` to drop the app's own server-driven `_state` echoes | Track the layouts you push and compare, or accept the echo. This version ships no native origin signal — the settled-`_state` gating from #93 removes *transients*, not *provenance*, so a settled echo of your own push is still indistinguishable from a user gesture. |

Fixes #92


tests/testthat/test-state-hygiene.R
```diff
@@ -72,28 +23,29 @@ test_that("_state tracks a settled container resize, tagged client", {
...
-  # Shiny activity until it fires, so poll for it rather than wait_for_idle.
+  # must reflect it -- consumers read the input directly. The debounced emit
+  # produces no Shiny activity until it fires, so poll for the settled width
+  # rather than wait_for_idle.
   app$set_window_size(width = 820, height = 700)
   for (i in 1:30) {
     Sys.sleep(0.1)
-    if (identical(app$get_value(input = "dock_state-source"), "client")) break
+    now <- grid_width()
+    if (now > 0 && !identical(now, before)) break
```
DivadNojnarg: **The rewritten poll exit can pass without the resize ever being exercised.**

`now > 0 && !identical(now, before)` is satisfiable by the *initial seeded* `_state` rather than by the resize emit. If the ResizeObserver's 150 ms debounce has not fired by the time `wait_for_idle()` returns -- plausible on slow CI -- then `before` is 0, the first poll iteration reads the seed width, breaks immediately, and both `expect_gt(after, 0)` / `expect_false(identical(after, before))` pass against a layout that was never resized.

The old `"client"` tag assertion was doing real work here: it was a signal only the resize path could produce. Cheapest restoration is an `expect_gt(before, 0)` right before `set_window_size()`, so the test fails loudly if it starts from an unseeded state instead of silently going green.

NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 - `move_panel()` now places the moved panel with the same `position` vocabulary as `add_panel()` and `panel()` — a list carrying a `referencePanel` or `referenceGroup` plus a `direction` (one of "above", "below", "left", "right", "within") and an optional `index` — replacing the previous scalar `position` / `group` / `index` arguments. Server-side callers can now express "move panel X next to panel Y" with the same placement grammar they use to add panels.
 
+- Removed `input[["<dock_ID>_state-source"]]` (added in 0.3.0) and the provenance machinery behind it. It tagged each `_state` update `"server"` or `"client"` so a consumer could ignore an echo of a layout it had just pushed; no consumer relies on it any more, and dockview-core v7 surfaces the same programmatic-vs-user origin natively through its layout-mutation transaction events, so the hand-rolled tag is not worth extending. A restore's settled `_state` — the one signal a consumer still wants — continues to arrive on its own at `onDidLayoutFromJSON`.
```
DivadNojnarg: **The stated rationale does not hold for the version this release ships.**

"dockview-core v7 surfaces the same programmatic-vs-user origin natively" is presented as present-tense justification, but the release pins and bundles 4.13.1, where no such provenance signal exists.

The adjacent settled-`_state` entry does not cover the gap either: it fixes *transients*, not *provenance*. A settled echo of a layout the consumer itself just pushed is still indistinguishable from a user gesture, so the original #70 use case (ignore your own echo) has no replacement in this version.

Removing the tag is still defensible if no consumer relies on it -- just worth scoping the sentence as forward-looking ("a future dockview-core will surface ... so the hand-rolled tag is not worth extending") so nobody reads it as an available alternative today.

DivadNojnarg (commented):
@nbenn This one is clean. Two low findings inline, neither blocking.

Verification I ran beyond reading the diff:

- Rebuilt the bundle from `srcjs` in a clean worktree with `dockview-core@4.13.1` and `npm run production`: output is **byte-identical** to the committed `inst/htmlwidgets/dockview.js`. The minified resize branch matches the new source (`e&&(r=!0),n(),i()` = `if (seeding) seeded = true; persistState(); refitWidgets();`).
- `git grep` across the branch: no remaining reference to `state-source`, `state_source`, `withServerDriven`, `runServerDriven` or `serverDrivenFor` anywhere in code, `R/`, `man/`, vignettes or examples. Every base-branch call site was removed; nothing dangles.
- The example-dir rename is a true rename with no other referrers, and there are no `_snaps/state_source*` snapshots left to go stale. All four changed R files parse.

The JS deletions are behaviourally equivalent modulo the retired tag: `withServerDriven` only set and cleared a flag and forwarded a return value that was never used, so unwrapping the eight handlers and the ResizeObserver seeding branch changes no ordering, no error propagation and no emit count. `_set-size` / `_set-panel-title` / `_restore-state` each still do exactly one explicit `saveDock`, and the `restoring` / `seeded` / `persisting` gates from #93 are untouched.

tests/testthat/test-state-hygiene.R
nbenn: Agreed, and thanks — that was a real hole. The old `"client"` assertion was the only thing tying the exit condition to the resize path, and `now > 0` doesn't replace it: an unseeded start makes `before` 0 and the first read satisfies the predicate against a layout nothing ever resized.

Added the `expect_gt(before, 0)` precondition you suggested, and since the baseline is now guaranteed non-zero I dropped the redundant `now > 0` from the loop so the exit is purely "the width changed":

```r
  # The poll below exits on a width that differs from this baseline, so the seed
  # has to have landed first: starting from an unseeded 0 the very first read
  # would satisfy it and the resize would never be exercised.
  before <- grid_width()
  expect_gt(before, 0)

  app$set_window_size(width = 820, height = 700)
  for (i in 1:30) {
    Sys.sleep(0.1)
    if (!identical(grid_width(), before)) break
  }
```

Worth noting the precondition holds deterministically here rather than by luck: this example's `output$state` reads `input[["dock_state"]]`, so the seed's flush is observable and `wait_for_idle()` returns after it. That is exactly the property the demo app lacked in #93, which is why `test-demo.R` needed an explicit `wait_for_value()`.

NEWS.md
nbenn: Fair, and the second half of your point is the sharper one: the settled-`_state` entry fixes transients, not provenance, so a consumer that used the tag to drop its own echo genuinely has no in-version replacement. Presenting v7 in the present tense papered over that.

Reworded to scope the upstream claim as forward-looking and to state the gap outright rather than leave it to be inferred:

> Removed `input[["<dock_ID>_state-source"]]` (added in 0.3.0) and the provenance machinery behind it. It tagged each `_state` update `"server"` or `"client"` so a consumer could ignore an echo of a layout it had just pushed. No consumer relies on it any more, and a future `dockview-core` will report the same programmatic-vs-user origin natively through its layout-mutation transaction events, so the hand-rolled tag is not worth carrying until then. Note that the bundled 4.13.1 carries no such native signal: an app that still needs to tell its own echo from a user gesture has to track the layouts it pushes itself, since the settled-`_state` gating above removes transients, not provenance.

I dropped the "a restore's settled `_state` continues to arrive at `onDidLayoutFromJSON`" sentence while I was in there — it was doing the same conflation in miniature, offering a transient-level guarantee as if it answered the provenance question.

PR #96, opened 2026-07-22
Emit a dockview:active-panel DOM event on activation

### LINKED ISSUE #95: Client-side hook when a panel's content mounts, for relocating DOM into it before paint
Consumers that relocate existing DOM into a panel — rather than handing content over as HTML — have no reliable client-side moment to do it, because a panel's content element is created lazily and nothing signals when it appears.

The content element carrying the id `<dockId>-<panelId>` is created in `Panel.init` ([`srcjs/modules/components.js`](https://github.com/cynkra/dockViewR/blob/3a0d43fdc5c7a664e84fcac8228fa9b0a28e1dd1/srcjs/modules/components.js#L17-L22)). dockview-core instantiates that content component only when the panel is visible — `DockviewComponent` resolves the renderer to `'onlyWhenVisible'` by default and dockViewR does not override it — so a group's non-front tab has no content element until it is first activated. Measured on a three-tab group whose middle tab was never activated: its `#<dockId>-block_panel-<middle>` element is absent from the DOM entirely, while the group's authored-active tab and the fronted tab both have theirs.

This bites any consumer that pushes DOM into the panel. In blockr.dock ([BristolMyersSquibb/blockr.dock#361](https://github.com/BristolMyersSquibb/blockr.dock/issues/361)), block cards live in an offcanvas and are moved into their panels with a Shiny custom message issued right after restore. For a background tab the move lands before the element exists and is dropped; the card strands in the offcanvas and the tab renders blank.

The signals that exist today do not cover the gap cleanly:

- The `_active-panel` input is a server round-trip. By the time the server learns a tab is active and sends the relocation back, dockview has already shown the empty panel, so the content pops in a frame later — a visible flash.
- A consumer-side `MutationObserver` catches the element client-side, before paint, but is a broad DOM watcher: it needs `subtree: true` to see the nested panel, so it also fires on unrelated mutations (embedded output re-renders) and lingers as long as a tab stays unopened.
- Forcing every tab to mount at restore, by cycling activations, is eager and reintroduces the settle jitter that the single-focus restore was meant to avoid.

There is already a client-side `onDidActivePanelChange` subscription in the widget ([`srcjs/modules/callbacks.js`](https://github.com/cynkra/dockViewR/blob/3a0d43fdc5c7a664e84fcac8228fa9b0a28e1dd1/srcjs/modules/callbacks.js#L96-L101)). Would surfacing a consumer hook there be worth considering — a callback, or a custom DOM event, fired when a panel's content mounts (and/or activates), carrying the panel id and its content element? A consumer could then relocate its DOM synchronously, before paint, with no server round-trip and no DOM watching. A first-class "render into a panel" API that owns the relocation would serve the same need at a higher level.

Happy to prototype whichever direction sounds right as a PR. This touches the same mount lifecycle as [#77](https://github.com/cynkra/dockViewR/issues/77).

nbenn:
## Summary

- Re-broadcasts `onDidActivePanelChange` as a bubbling `dockview:active-panel` DOM event on the dock container, carrying the active panel id in `event.detail.id` and the dock id in `event.detail.dock`.
- The widget's dockview `api` is closure-private, so this event is the only client-side handle a consumer's JavaScript has onto activation — the counterpart to `input[["<dock_ID>_active-panel"]]` without the server round-trip. A consumer relocates deferred DOM into a panel on the tick its content mounts (dockview mounts a background tab's content lazily on first activation), before paint and with no document-wide observer — the need in BristolMyersSquibb/blockr.dock#361.
- Narrowed per the issue discussion: the "relocate DOM into a panel" API is dropped — that policy stays with the consumer; dockViewR only surfaces the signal.
- `_active-panel` is written before the dispatch, so a listener that activates another panel re-enters the handler and its nested write wins rather than being overwritten by a superseded id.
- Rebuilt the JS bundle on the pinned dockview-core 4.13.1; regression tests cover both the re-broadcast and the re-entrant ordering (each proven to fail against the unfixed code), plus a NEWS entry documenting the payload and the bind-before-render constraint.

Fixes #95


srcjs/modules/callbacks.js
```diff
@@ -97,7 +97,24 @@ const setDockViewCallbacks = (id, api) => {
...
+
+    if (e === undefined) return;
+
+    // Re-broadcast activation as a DOM event on the container: the dockview `api`
+    // is closure-private, so this is a consumer's only client-side handle onto
+    // activation. A consumer listens to relocate deferred DOM into a panel on the
+    // tick it mounts -- before paint, no server round-trip.
+    const container = document.getElementById(id);
+    if (container) {
+      container.dispatchEvent(
```
DivadNojnarg, line 156: **The DOM event is dispatched before `_active-panel` is set, so a re-entrant listener leaves the input stale.**

`dispatchEvent` is synchronous. A listener that reacts by activating another panel (a reasonable thing to do -- "if panel X mounts, focus Y") re-enters `onDidActivePanelChange` inside this frame: the nested handler dispatches, then writes the *correct* new id to `_active-panel`, and only then does the outer frame resume and overwrite it with the now-superseded id.

Net effect: `input$<id>_active-panel` reports the panel that was activated first, not the one that is actually active, with no further event to correct it.

Setting the input before dispatching makes the outer write happen first and the nested write win, which is the correct ordering.

srcjs/modules/callbacks.js
```diff
@@ -97,7 +97,24 @@ const setDockViewCallbacks = (id, api) => {
...
+        new CustomEvent('dockview:active-panel', {
+          bubbles: true,
+          detail: { id: e.id }
```
DivadNojnarg: **`detail` omits the dock id, but NEWS recommends binding on `document`.**

Panel ids are only unique *per widget* -- `check_panel_ids()` scopes the uniqueness check to a single `dock_view()` -- so two docks on one page can both own a panel called `"main"`. A `document`-level listener receiving `detail: { id: "main" }` cannot tell which dock fired, nor whether to target `#left-main` or `#right-main`.

The only disambiguator today is `event.target.id`, which is neither documented in the NEWS entry nor covered by the new test. Adding the dock id to `detail` (`detail: { id: e.id, dock: id }`) makes the documented binding pattern actually sufficient, and costs nothing.

srcjs/modules/callbacks.js
```diff
@@ -97,7 +97,24 @@ const setDockViewCallbacks = (id, api) => {
 
   api.onDidActivePanelChange((e) => {
     requestSync();
-    if (HTMLWidgets.shinyMode && e !== undefined) {
+
+    if (e === undefined) return;
+
+    // Re-broadcast activation as a DOM event on the container: the dockview `api`
```
DivadNojnarg, line 149: **A listener bound after `renderValue()` never learns about the already-active panel.**

The first activation fires inside `renderValue()`, before `_initialized`, and the `api` is closure-private -- so there is no way to query current activation after the fact. Any consumer that binds later misses it permanently and its deferred content is silently never relocated. Two ordinary cases hit this:

- a dock rendered inside `renderUI()` / `insertUI()`, where the consumer's JS runs after the widget has already rendered;
- a consumer that (sensibly) defers binding until `_initialized`, exactly to be sure the dock exists.

The new example only works because it binds in `tags$head` at page parse time, which reads as the pattern working generally when it is in fact the one ordering that does.

Worth either re-dispatching the current activation on `_initialized`, or documenting the constraint explicitly in NEWS so consumers know binding must precede render.

srcjs/modules/callbacks.js
```diff
@@ -97,7 +97,24 @@ const setDockViewCallbacks = (id, api) => {
...
+    // is closure-private, so this is a consumer's only client-side handle onto
+    // activation. A consumer listens to relocate deferred DOM into a panel on the
+    // tick it mounts -- before paint, no server round-trip.
+    const container = document.getElementById(id);
```
DivadNojnarg: **Redundant re-query.** `document.getElementById(id)` shadows the container already resolved in the outer scope (line 129) and re-queries the DOM on every activation. Reusing the outer binding drops the lookup and the `if (container)` divergence between the two call sites.

DivadNojnarg (commented):
@nbenn Small diff, and the central premise checks out. Four findings inline, the first two worth addressing before merge.

Verified as correct: the bundle matches the source change exactly (single dispatch site, esbuild-mangled equivalently); callbacks are wired *before* `initDockPanels()`, so the initial activation of panel "a" is observable and the test's first assertion genuinely holds; dockview's `doSetActivePanel` appends the panel content element (`contentContainer.openPanel`) *before* firing `_onDidActivePanelChange`, so the PR's central claim -- the element exists on the tick the event fires, before paint -- is true; `panel(active = FALSE)` maps to `inactive: true` -> `skipSetActive`, so "b" does not activate at init; the `select_panel()` / `panel()` signatures in the new example are right; and the default `mode = "prod"` means `evalDockView` actually runs the callback. A throwing consumer listener also cannot break dockview's emitter, since the browser swallows DOM listener errors.

srcjs/modules/callbacks.js
nbenn: Confirmed and fixed in 171da7c — the input write now precedes the dispatch.

I reproduced it end-to-end rather than reasoning about it: a listener that activates another panel on `dockview:active-panel` (dispatching `pointerdown` on the sibling tab, which is a consumer's client-side route to activation since `api` is private). With the old ordering `input$dock_active-panel` settles on `"b"` while `"c"` is the active panel; with the write first it settles on `"c"`. That is now a regression test (`test-active-panel.R`, "a re-entrant listener leaves the active-panel input current"), and I checked it fails against the pre-fix ordering — the DOM-event assertion passes either way, so the input assertion is the one carrying the ordering contract.

srcjs/modules/callbacks.js
nbenn: Agreed — fixed in 171da7c. `detail` is now `{ id, dock }`.

You are right that the NEWS entry was writing a cheque `detail` could not cash: recommending a `document`-level listener while shipping only a panel id that is unique per widget. NEWS now documents `event.detail.dock` alongside `event.detail.id` and says explicitly that a listener bound above one dock needs it to tell which fired. The test asserts the payload (`event_dock`), and the re-entrancy test resolves its target tab through `e.detail.dock`, so the disambiguator is exercised as the way to get from an event back to the right dock rather than just asserted as a literal.

srcjs/modules/callbacks.js
nbenn: Went and checked this against the consumer that motivated the issue rather than leaving it a judgement call — blockr.dock's relocation handler has since merged, and it settles what the gap actually costs.

Its listener, from `inst/assets/js/show-block.js`:

```js
$(function () {
  var pending = [];
  // ...
  document.addEventListener('dockview:active-panel', () => {
    pending = pending.filter((m) => alive(m) && !applyMove(m));
  })
})
```

Three properties decide it. It binds inside a DOM-ready handler in a package asset, so the binding strictly precedes any dock's render — including docks inserted later by `renderUI()` / `insertUI()`, the case I was least sure of. It ignores `event.detail` entirely, using the event as a "something mounted, retry the stash" pulse and re-testing each pending move's target selector; that is idempotent retry rather than edge-correlated action, so it survives a missed edge as long as any later activation occurs. And the one situation where the initial activation genuinely carries weight — a `move-element` arriving before the dock has rendered, so even the front panel's element is missing — is covered precisely because the binding predates the render.

So for the motivating consumer the gap is immaterial. That is not an argument that the sharp edge isn't real: it is exactly as you describe for anyone binding later, which is why NEWS now states that the event reports transitions only, that a page-load `document` listener sees every dock including later-inserted ones, and that binding after render (from `_initialized`, say) misses that dock's initial activation with no way to query it afterwards.

We are merging on the documented behaviour, so I am resolving this — on that basis, not because the design question is closed. A replay on `_initialized` would not disturb blockr.dock either, since one extra idempotent retry is a no-op for it. If you would still rather the event be self-sufficient than documented, say so and I will open a follow-up.

srcjs/modules/callbacks.js
nbenn: Fixed in 171da7c — hoisted `const container = document.getElementById(id)` to the top of `setDockViewCallbacks`, so the handler and the pointerup / ResizeObserver block share one lookup and the shadow is gone.

Hoisting rather than referencing the later binding from the callback, since that would have left the handler depending on a `const` declared below it — legal only because the first activation happens during `initDockPanels()`, after this function returns. That is exactly the kind of TDZ-adjacent ordering assumption worth not encoding.

srcjs/modules/callbacks.js
nbenn: Data point after the fact, since it validates the catch: blockr.dock consumes this input server-side — `board-server.R` feeds `input[["<dock_id>_active-panel"]]` into `visible_block_ids(layout, dock$active_panel())`, which drives its render/visibility gate. A stale `_active-panel` would have marked the wrong block visible, the same failure mode as the downstream issue this event exists to fix. So the re-entrancy hazard had a live consumer waiting for it rather than a hypothetical one.

Issue #99, opened 2026-08-12
Bundle grows from 214 KB to 500 KB after the dockview v8 upgrade


DivadNojnarg:
Not urgent, filed for later. Nothing is broken; this is page weight.

The dockview v8 upgrade takes `inst/htmlwidgets/dockview.js` from 214 KB to 500 KB. Breakdown from the webpack build:

- ~150 KB of that is `dockview/dist/styles/dockview.css`, up from ~42 KB at 4.13.1. The growth is mostly the theme set going from 8 to 18, plus styles for v8 features we do not ship (pinned tabs, tab group chips, DnD compass and so on, all commercially licensed upstream).
- The remainder is genuinely new v8 JavaScript: the module system, edge groups, tab groups, per-window live regions.

Tree-shaking is not the problem. Upstream sets both `module` (`./dist/package/main.esm.mjs`) and `sideEffects: ["**/*.css", "**/*.scss"]`, and webpack resolves the ESM entry, so unused JS exports are already being dropped.

The CSS is the tractable half, and the reason it is not tree-shaken is that we import the compiled stylesheet wholesale in `srcjs/widgets/dockview.js`:

```js
import 'dockview/dist/styles/dockview.css';
```

Options, roughly in order of effort:

1. Import only the theme(s) actually requested. Awkward, because `theme` is a runtime argument to `dock_view()` and the bundle is built once, so this would mean shipping all themes as separate lazily-loaded chunks rather than one stylesheet.
2. Build the CSS from upstream's Sass sources instead of the compiled bundle, importing only the theme partials we expose. `sass-loader` is already a devDependency. Needs a look at whether upstream ships the partials in the published package.
3. Drop the `-spaced` theme variants, halving the theme count. Breaking, and probably not worth it for bytes.
4. Do nothing. 500 KB is well inside CRAN limits and is a one-time cost per session, cached thereafter.

Option 4 is a legitimate answer. Worth measuring the actual first-paint impact before spending effort on 1 or 2.

Context: #97, and the upgrade itself.

Issue #97, opened 2026-08-11
Upgrade bundled dockview to v8


DivadNojnarg:
`main` pins `dockview-core@^4.13.1`; upstream latest is `8.0.0`, so the gap spans four majors. The WIP branch `chore-dockview-v7` already walked 4 -> 6 -> 7 but is 45 commits behind `main` with a 2025-12-09 merge base. #66 (edge groups) targets that branch and itself branched at the v6.3.0 commit, so it has never seen the 6.6.1 or 7.0.2 bumps.

Findings below are from reading both upstream migration guides (only v7 and v8 exist) and from inspecting the published 4.13.1 and 8.0.0 packages side by side.

## What actually breaks

Cross-referenced the dockview API surface `main/srcjs` uses against both guides.

| Change | Affects us |
| --- | --- |
| **`dockview-core@8` ships no stylesheet.** `dist/styles/dockview.css` is gone from the package `files`. `srcjs/widgets/dockview.js:2` imports exactly that path, so the webpack build fails outright. The CSS now lives in `dockview/dist/styles/dockview.css`. | Yes. This makes the `dockview-core` -> `dockview` package switch **mandatory**, not the "recommended" the v7 guide frames it as |
| v7: `onDidActivePanelChange` emits `{ panel, origin }` instead of the panel | Yes. `srcjs/modules/callbacks.js` needs `e.panel?.id` |
| v6 era: theme exports changed, `themeReplit` gone, 18 themes now | Yes, and already done on `chore-dockview-v7`. Verified v8 exports exactly those 18 symbols, so `themes.js` ports verbatim |
| v8: panel dimensions report content area, not group box | No. We never call `onDidDimensionsChange` |
| v8: `moveToNext`/`moveToPrevious` -> `activateNext`/`activatePrevious` | No. Unused, and the old names survive as deprecated aliases |
| v7: `rootOverlayModel`, `onUnhandledDragOverEvent`, assorted type renames | No. All unused |

## Serialised layout compatibility: no migration needed

`SerializedDockview` in 8.0.0 is 4.13.1's shape plus an optional `edgeGroups?`; the nested `GroupPanelViewState` gains an optional `tabGroups?`. Both optional, no version field, and no migration branch in `fromJSON`. A layout serialised by 4.13.1 is valid v8 input, so stored `_state` in downstream consumers survives the bump. Only the reverse breaks, a v8 layout using edge or tab groups will not load in 4.x, which matters only on a downgrade.

## Licence boundary

[...]

PR #100, opened 2026-08-12
Upgrade bundled dockview to v8

DivadNojnarg:
## Summary

Moves the bundled dockview from `dockview-core` 4.13.1 to `dockview` 8.0.0, four majors of upstream fixes. Despite the version span, the surface that actually breaks is small: I enumerated every dockview API `srcjs` touches and cross-referenced it against upstream's v7 and v8 migration guides (the only two that exist).

Three things forced changes.

**The stylesheet moved packages.** `dockview-core@8` ships no CSS at all: `dist/styles/dockview.css` is gone from the package's `files`. Since `srcjs/widgets/dockview.js` imports exactly that path, the build fails outright until the dependency becomes `dockview`, which is where the stylesheet now lives. This is why the package name changes and not just the version. Upstream's v7 guide frames the `dockview-core` -> `dockview` switch as merely recommended; for us it is mandatory.

**`onDidActivePanelChange` changed shape.** It now emits `{ panel, origin }` rather than the panel itself, and `panel` is `undefined` when a dock is left with no active panel. The handler reads `e?.panel` and guards on it. This is the only place the v7/v8 payload changes bite us: I checked every other subscribed event against the v8 types and `onDidActiveGroupChange`, `onDidAddPanel`, `onDidRemovePanel`, `onDidAddGroup`, `onDidRemoveGroup`, `onDidMovePanel`, `onDidLayoutFromJSON`, `onDidMaximizedGroupChange` and `onDidTitleChange` all keep their existing payloads.

**Upstream dropped the Replit theme** and added eleven others, taking the total from 8 to 18. `matchTheme()` becomes a lookup rather than a nine-arm switch, and the five examples that used `replit` move to the default `light-spaced`.

v8's two documented breaking changes do not affect us: we never call `onDidDimensionsChange`, so the content-area dimension change is moot, and we never call `moveToNext`/`moveToPrevious`, which are renamed but kept as deprecated aliases anyway.

## Serialised layouts still restore

Worth calling out because it was the main risk going in. dockview 8's `SerializedDockview` is 4.13.1's shape plus an optional `edgeGroups?`, and the nested `GroupPanelViewState` gains an optional `tabGroups?`. Both optional, no version field, no migration branch in `fromJSON`. A stored `input[["<dock_ID>_state"]]` from the previous version needs no migration.

The test suite corroborates this independently: across every snapshot in the suite, the only JSON changes are the two `theme: "replit"` -> `"light-spaced"` strings from the example edits. Zero geometry drift, zero structural drift. Screenshots shift by a few pixels of tab padding and a redrawn maximize glyph.

## Licence file now ships

`inst/htmlwidgets/dockview.js.LICENSE.txt` is new and tracked. Webpack extracts the MIT notice out of the minified bundle, so without committing it the shipped bundle would carry no attribution, which MIT requires.

## Only the MIT core is bundled

v8 splits its feature set across an MIT core and a separately licensed commercial package (`dockview-enterprise`). We bundle only the MIT core. The boundary is verifiable in the shipped code: `moduleContracts.d.ts` exports 33 interfaces with zero implementations, and those are exactly the paid features.

The practical consequence is a footgun worth knowing about: because `dock_view()` forwards arbitrary `...` to `createDockview`, and v8's types accept the commercial option names whether or not an implementation is registered, `dock_view(..., smartGuides = TRUE)` is accepted and silently does nothing. NEWS documents this. Whether to validate `...` against the free option set and warn is left as a separate decision rather than settled here.

## Verification

- Full `testthat` suite green: all 7 files, including the 9 `state-hygiene` apps and the 9 `proxy` apps. The settled-`_state` gating from #93, the `_state-source` retirement from #94 and the `dockview:active-panel` event from #96 all survive the upgrade untouched.
- `R CMD check`: 0 errors, 0 warnings, 2 NOTEs. Both are unrelated to this change (the standard "unable to verify current time", and a local gitignored `CLAUDE.local.md` at top level).
- Note on coverage: check's test stage runs in 81s because the 19 `skip_on_cran()` gates skip every `AppDriver` test under check. The browser tests are covered by the `test_local()` run above, not by check.

## Deliberately not included

The `chore-dockview-v7` branch also carried an `_active-group` staleness fix (a capture-phase `pointerdown` handler in `components.js`), a new `active_group` example, and a regression test for it. Those three are one self-contained bug fix with no dependency on the version bump, so they are left out to keep this reviewable. `api.isActive` and `api.group` both still exist in v8, so it ports cleanly whenever it lands.

Follow-ups: #99 (bundle grows from 214 KB to 500 KB, filed for later, not a regression to fix here).

Refs #97. Supersedes the stale `chore-dockview-v7` branch, which was 45 commits behind `main`; #66 will be redone on top of this rather than rebased.


christophsax:
Ran this against a real board — CEDX, 9 tables / 246K rows, 11 views, crossfilter, patient profile, mini-DAG — on top of `blockr.core@main` and `blockr.dock@main`. Boots clean, view switches restore layouts, deferred panels construct on tab activation, save→reload round-trips pixel-identical. `blockr.dock`'s suite is 7805 pass / 0 fail / 0 skip against v8. No dockview errors in console. Happy to merge.

### Merge `main` first, and rebuild the bundle rather than resolving by hand

The branch is 5 commits behind, and those 5 are #101's scoped-rebind work (bind panel bodies once, rebind on return) — which is what keeps inputs live in hidden panels, so it matters more on a big board than anywhere else. Only `inst/htmlwidgets/dockview.js` conflicts; `npm ci && npm run production` resolves it, 500 KiB as predicted.

### The "silently does nothing" footgun is already fixed upstream

v8 ships `OPTION_MODULE_RULES`, which validates options at construction and on every `updateOptions` and logs one deduplicated console error per missing module, naming the exact option path:

```
dockview: `smartGuides` requires the "SmartGuides" module, which ships in dockview-enterprise.

  npm install dockview-enterprise
  import 'dockview-enterprise'; // self-registers every enterprise module
```

I set three paid options at once and got three precise errors. So the NEWS wording should change, and the open question about whether to validate `...` against the free option set is answered — upstream does it for us.

### The serialisation section is arguing a risk we don't carry

`blockr.dock` persists its own `dock_grid` (orientation / children / sizes / active, no volatile ids) and re-derives dockview's layout via `as_dock_layout()`. I checked a saved CEDX project: zero dockview keys in a 112 KB payload. `dock-grid.R` says it outright — "swapping dockView touches only `dock_layout` and those two casts". The compatibility claim is right, and stronger than the PR argues.

### Worth setting expectations on what v8 buys an MIT consumer

All six headline features in dockview's "what's new in 8.0" — pinned/wrapping tabs, drop compass, floating-group snapping, layout history, auto-hide edge groups, keyboard docking — are on the enterprise list. What the free core gains is the module architecture, edge groups, and the missing-module diagnostic. That's a fine reason to merge; it just isn't "four majors of fixes" from where we sit.

### Follow-up worth filing: edge groups

These are free (dockview's own licence page lists them under MIT) and they work — I put the filter panels of the CEDX board on a left rail as vertical tabs and they switch correctly:

```js
var grp = api.addEdgeGroup('left', { size: 320 });
var target = api.getGroup(grp.id);
ids.forEach(id => api.getPanel(id).api.moveTo({ group: target, position: 'center' }));
```

Getting there properly needs an `add_edge_group()` proxy in dockViewR and `dock_grid` learning about edges in blockr.dock, since a rail doesn't survive a save today. Scope it as a static rail: I verified that click-to-collapse is the enterprise `AutoHideEdgeGroup` module (`pinEdgeGroup`, `autoHideEdgeGroup` and `peekEdgeGroup` all log the missing-module error), and neither resizing the group nor `setEdgeGroupVisible()` substitutes for it.

PR #101, opened 2026-08-14
Bind panel bodies once rather than on every layout sync

### LINKED ISSUE #98: Every layout sync rebinds all panels and dispatches a global window resize
## What

Every dockview event funnels through `requestSync()` in `srcjs/modules/callbacks.js`, which runs two whole-page side effects regardless of what the gesture actually changed:

- `persistState()` (`callbacks.js:24-33`) loops `api.panels` and calls `Shiny.initializeInputs($(pane))` + `Shiny.bindAll($(pane))` for every panel in the dock.
- `refitWidgets()` (`callbacks.js:15-17`) dispatches a global `window` `resize` event.

Neither is scoped to what moved. A pure-geometry gesture — a sash drag, a maximize toggle, a panel reordered inside its group — leaves every panel body untouched, and still pays both.

## Why each one is expensive

Both are amplified by Shiny into whole-page work (line numbers from shiny 1.14 `www/shared/shiny.js`):

- `Shiny.bindAll()`'s `bindOutputs()` ends with an unconditional `setTimeout(() => sendOutputInfoFns.regular(), 0)` (`:5842`; `unbindOutputs()` the same at `:5900`), and `doSendOutputInfo()` (`:7569`) walks **every** `.shiny-bound-output` on the page, re-sending `.clientdata_output_<id>_hidden` plus `_width` / `_height` for size-reporting outputs. The `scope` argument narrows the binding, not the refresh. The already-bound fast path (`:5822`) skips the element but not the scheduled refresh, so re-binding panels whose DOM did not change still costs the full walk.
- A global `resize` is broader still: it drives every bound output's `ResizeObserver` (`:7537`), Shiny's own theme-refresh signal (`registerThemeRefreshSignals`, `:7506-7517`, which walks `.shiny-bound-output` a second time), and a `.clientdata_pixelratio` write (`:7589`).

One clarification worth recording, because it changes what the fix should be: the N per-panel `bindAll` calls inside a single `persistState()` do **not** cost N refreshes. `sendOutputInfoFns.regular()` goes through a 0ms `Debouncer` (`:352`), so they collapse to one whole-page refresh per sync. **The cost scales with the number of syncs, not the number of panels.** Narrowing the loop's scope therefore buys nothing on its own — the win is in not running the sync at all when no panel DOM changed.

## Measured

Instrumented by wrapping `Shiny.bindAll` / `unbindAll` / `initializeInputs` in the page and counting calls, plus `.clientdata_output_*` input writes, against a blockr.dock board of 8 blocks + 1 extension panel (45 bound outputs).

Per gesture, after the app has settled:

| gesture | `bindAll` | `unbindAll` | `initializeInputs` | `.clientdata_output_*` writes | server round-trips |
| --- | --- | --- | --- | --- | --- |
| window resize | 9 | 0 | 9 | 67 | 2 |

Nine `bindAll` + nine `initializeInputs` is one `persistState()` over the dock's nine panels, and the 67 clientdata writes are the whole-page walk it plus `refitWidgets()`' synthetic `resize` provoke — on a page with 45 bound outputs, for a gesture that changed no panel's DOM at all.

The container-resize branch (`callbacks.js:202-226`) is the one I could drive headlessly: I could not synthesise a sash drag or a tab activation that dockview accepted from a scripted event, so I am not claiming numbers for those. They reach the same two functions through `requestSync()` — `onDidMovePanel` (`:75`), `onDidMaximizedGroupChange` (`:104`), `onDidActivePanelChange` (`:136`), `onDidActiveGroupChange` (`:165`) and the sash `pointerup` (`:183-188`) — so the shape should be the same, but that part is read off the source rather than measured.

For contrast, at **mount** these sites are nearly silent: of the 85 bind / unbind / initializeInputs calls the same board makes while starting up, exactly **1** comes from `dockview.js`. The restore guards (`isRestoring`, `persisting`, `flushScheduled`) suppress the rest, and the mount's real amplifier is ordinary Shiny output rendering. So this is a gesture-time cost, not a startup cost — worth stating up front, since it was originally filed the other way round in BristolMyersSquibb/blockr.dock#397.

## Candidate work

1. **Only persist / rebind when a panel's DOM actually changed.** `onDidMovePanel`, the sash `pointerup` and `onDidMaximizedGroupChange` are geometry-only; the `_state` publish still has to happen, the `Shiny.bindAll` loop does not. `onDidLayoutFromJSON` and `onDidAddPanel` are the cases that genuinely need a rebind.
2. **Scope or drop `refitWidgets()`'s global `window` resize.** dockview knows which panels changed size; notifying those (or letting each panel's own `ResizeObserver` do it) avoids the whole-page theme walk and the pixelratio write that a synthetic window resize triggers.
3. **Gate the per-panel inputs no consumer reads.** `_added-panel`, `_removed-panel`, `_n-panels` and `_n-groups` are written on every panel / group add and remove. blockr.dock reads none of them — it reads `active-panel`, `active-group`, `initialized`, `panel-to-remove`, `panel-to-add` and `state`, and derives its panel count from a server-side membership set instead (`R/board-server.R:697`). `_restored` is unread too. They ride in the same input batch as the consumed ones, so this costs no extra round-trip: it is reactive churn and wire noise, not latency. Cheapest of the three, and the least valuable.

(1) is the concrete form of #72. (2) and (3) are independent of it. #77 would subsume part of (1) if the surgical apply lands first.

## Reproducing

The probe is a page-side wrapper around the three `Shiny` entry points plus a `shiny:inputchanged` listener, driven from shinytest2 with `wait = FALSE` so it attaches before the app settles. Happy to attach the script if useful.

nbenn:
## Summary

- Panel bodies are now bound when they enter the document and again whenever they return to it, instead of on every layout sync. Each `Shiny.bindAll()` ends by scheduling a walk over every bound output on the page — re-sending each one's hidden state and, where it reports one, its size — and schedules it whether or not the scope it was handed holds anything unbound. Binding all panels on every sync therefore charged that walk to gestures that change no panel body at all: a sash drag, a maximize, a window or container resize.
- The condition to skip on is attachment, not "bound once". A background tab's body is removed from the document rather than hidden, and the walk a bind books is also how Shiny learns an output is on screen again — so a returning body has to be rebound or a tab switch gets slower. Binding once per element measured 151ms to un-hide a re-revealed output against 17ms on main; keying on attachment brings that back to 30ms while skipping the same geometry-only gestures. The record holds the element rather than a flag, which also covers a restore rebuilding bodies under the ids they already had.
- There is no event allowlist to keep in sync with dockview: `bindPanels()` rescans on every sync, so a body that reaches the document without an event of its own is still picked up by the next one.
- Three tests. A gesture that changes no panel body must book no bind (sash release and container resize, with adding a panel as the control that keeps the assertion from passing vacuously); a background tab brought to the front for the first time must come back with working inputs; and a tab hidden and shown again must rebind exactly its own pane — two calls, not the six that rebinding the whole dock would cost — and still report its slider.

## Measured

The board the issue measured — a `blockr.dock` board of 8 blocks plus the edit extension, 45 bound outputs — counting `Shiny` entry-point calls and `.clientdata_output_*` input writes per window resize. Six samples per build, every one identical.

| per window resize | `bindAll` | `initializeInputs` | `.clientdata_output_*` | busy periods |
| --- | --- | --- | --- | --- |
| before | 9 | 9 | 66 | 2 |
| after | 0 | 0 | 15 | 2 |

The before row reproduces the reading in the issue (9 / 9 / 67 / 2). What remains afterwards is the per-output size report for the outputs that genuinely changed size, which is the work a resize is supposed to do. The round-trip count is unchanged, so this removes client work and wire noise rather than latency.

Tab switching was measured separately, being the interaction this could have made worse, and is unchanged: 30ms to un-hide a re-revealed output against 17ms before. That same board also reports the same 30 cards, 45 page-wide bound outputs and 47 bound inputs inside panels as before the change, across mount, a resize and a sash release — its cards reach the panels through blockr.dock's own `move-element` handler, which is the path most likely to notice a binding change.

## The issue's other two candidates, and why neither becomes code

The global `window` resize is not the expensive half. Dispatched on a settled page it costs 0 client-data writes and 0 round-trips, four times out of four: Shiny's two window-resize handlers (theme refresh, pixel ratio) dedupe on unchanged values, and every `.shiny-bound-output` already carries its own `ResizeObserver` reporting size per element. The whole-page walk the issue attributed partly to it was booked by the rebind, which this PR removes. It also remains load-bearing for static htmlwidgets, which listen on `window` only, unlike Shiny-bound outputs.

Gating the per-panel inputs is declined rather than deferred. All five are advertised API with their own NEWS entries — `_added-panel` and `_removed-panel` shipped in 0.2.0, `_n-panels`, `_n-groups` and `_restored` in 0.3.0, and `_restored`'s firing order was tightened again in the current development cycle. Removing them breaks any app that took those entries at their word, and "blockr.dock reads none of them" is a fact about one consumer of a general-purpose widget package rather than about all of them. By the issue's own account they ride an existing input batch, so there is no latency to reclaim either. Should they ever go, that is a deprecation cycle of its own.

Fixes #98


nbenn:
@DivadNojnarg another performance improvement. What do you think?

srcjs/modules/callbacks.js
```diff
@@ -16,20 +16,64 @@ const setDockViewCallbacks = (id, api) => {
...
+
+    api.panels.forEach((panel) => {
+      live.add(panel.id);
+
+      const pane = document.getElementById(`${id}-${panel.id}`);
+      if (!pane) {
+        attachedPanes.delete(panel.id);
+        return;
+      }
+      if (attachedPanes.get(panel.id) === pane) return;
```
DivadNojnarg, line 55: **Unbound markup inserted into an already-attached panel body is never bound.**

This early return fires whenever a body has not moved, and nothing else invalidates the map entry, so a body that never leaves the document is never rebound after its first bind. Two single-panel groups keep both bodies attached permanently, which makes that the steady state rather than an edge case.

The concrete loser is a consumer using the `dockview:active-panel` hook this same file documents, to relocate deferred markup into the panel that just activated. The bind that used to pick that markup up was the per-sync rebind on `main`; now the sync that follows the activation returns here and the relocated inputs stay dead.

I am flagging this as a decision rather than a defect, because the behaviour it removes was never documented and your measurements say blockr.dock's `move-element` path is unchanged, so the in-tree consumer looks fine. But it is a real narrowing of what a consumer can rely on. Either accept it and say so in the NEWS entry, so anyone relocating DOM into a live panel knows binding is now their problem, or give them a way to ask for a rebind. Silently narrowing it is the option I would avoid, since the failure is dead inputs with no error.

NEWS.md
```diff
@@ -16,6 +16,8 @@
 
 ## Bug fixes
 
+- A layout gesture no longer rebinds every panel in the dock. Each `Shiny.bindAll()` ends by scheduling a walk over every bound output on the page — re-sending each one's hidden state and, where it reports one, its size — and schedules it whether or not the scope it was handed holds anything unbound. Binding all panels on every sync therefore charged that walk to gestures that change no panel body at all: a sash drag, a maximize, a window or container resize. Panel bodies are now bound once, on the first sync that finds one in the document — when a panel is added, when a background tab is brought to the front for the first time, and when a restore rebuilds the bodies. Measured on a nine-panel dock holding 46 bound outputs, a window resize went from nine `bindAll()` and nine `initializeInputs()` calls to none, and from 76–101 client-data input writes to 10–35, the remainder being the per-output size reports for the outputs that genuinely changed size.
```
DivadNojnarg: **This entry describes the earlier bind-once design, not the shipped one.**

It says a body is bound once, when "brought to the front for the first time". The shipped code keys on attachment and rebinds on **every** reattachment, which the new test asserts directly by expecting exactly 2 calls on re-reveal. As written the entry promises less than the code delivers, and would mislead anyone reasoning about whether a tab switch rebinds.

The measurements look like they come from that iteration too: the entry cites a nine-panel dock, 46 outputs and 76-101 -> 10-35 writes, while the PR body's table reports 45 outputs and 66 -> 15.

DivadNojnarg (commented):
@nbenn The mechanism holds up and the reasoning in the comment block is right. Two findings inline: one behavioural question worth a deliberate decision, one stale NEWS entry.

What I verified rather than assumed:

- The default `onlyWhenVisible` renderer does `element.remove()` and later re-appends **the same element object**, so element identity plus `getElementById` presence is a valid attach/detach oracle. `closePanel` confirms it.
- Maximize hides views via `classList.toggle("visible")` plus zero sizing rather than detaching or `display:none`, so "a maximize books no bind" holds, and it does not strand Shiny's hidden-state bookkeeping.
- The delete-during-`Map.forEach` cleanup is safe, panels dropped from `api.panels` are reaped, and a restore is covered because `fromJSON` builds fresh elements under the ids they already had.
- Moving from `$('#id-panelId')` to `getElementById` is incidentally more robust for ids containing CSS-special characters.
- The bundle matches the source change, and the three new tests reference real panel ids and inputs in the `state` / `serialise` examples (`state` has two groups, so the `.dv-sash` query resolves).

Two things I chased and decided not to report, in case you want them on record. `attachedPanes.set()` runs before the bind calls, so a throwing `Shiny.bindAll` would permanently mark that panel bound. I could not construct a case where the old per-sync retry would have succeeded, since the throw is deterministic and recurs on every attempt, so there is no behavioural differential, only a robustness smell. And the missing `Shiny.unbindAll` on detach plus the duplicate `addCustomMessageHandler` registration per re-render are both pre-existing and outside this diff.

srcjs/modules/callbacks.js
nbenn: Confirmed, and taking the first option: accepted, and documented.

I reproduced the boundary before deciding, injecting an unbound `action-button` into a panel body and then driving a sync:

| markup inserted into | main | this PR |
| --- | --- | --- |
| a body already in the document, which stays put | bound | **not bound** |
| a body entering the document, on the activation tick | bound | bound |

So the narrowing is exactly as you describe, and the `dockview:active-panel` pattern this file documents — relocating deferred content onto the tick a panel activates — sits on the right side of it, because that body is arriving and gets bound together with whatever it now holds.

The NEWS entry now gives the consequence a paragraph of its own, since nothing reports it when it bites: markup inserted into a body that is already in the document is the caller's to bind, with the `Shiny.initializeInputs()` / `Shiny.bindAll()` recipe spelled out and the mount-tick case called out as unaffected. There is also a test pinning the half of that boundary the entry promises, so the guarantee is not resting on prose alone.

I did not add a rebind hook, on the grounds that Shiny already is one: calling `Shiny.bindAll(el)` over the inserted subtree is what any code injecting markup into a Shiny page has to do anyway, and it is what a dockViewR-specific wrapper would forward to. Adding surface that only forwards seemed worse than documenting the standard move. Say the word if you would rather have an explicit handle and I will add one.

On the two you chased and set aside, I agree with both. Setting `attachedPanes` before the bind calls is the ordering I would keep, in fact: moving it after would retry a deterministic throw on every subsequent sync rather than once. The missing `unbindAll` on detach and the duplicated `addCustomMessageHandler` registration are pre-existing, and I have left them alone.

NEWS.md
nbenn: Fixed in bd577a7. The entry now describes what ships — bound on entry to the document and again on every return, left alone while it stays put — and carries the PR body's measurements: a nine-panel dock holding 45 bound outputs, nine `bindAll()` and nine `initializeInputs()` calls going to none, and 66 client-data writes to 15.

Good catch on the provenance of those numbers. They came from a synthetic nine-panel app I used while iterating; I replaced them in the PR body once I rebuilt the measurement on a real board and did not carry that back to NEWS.

PR #104, opened 2026-08-14
Add edge groups

DivadNojnarg:
## Summary

Adds edge groups: groups pinned to one edge of the dock, the usual shape for a filter rail or a file tree.

```r
dock_view(
  panels = list(
    panel(id = "main", title = "Main", content = "Main panel"),
    panel(
      id = "tree", title = "Tree", content = "Lives in the rail",
      # Naming the group is the whole instruction, so no `direction`.
      position = list(referenceGroup = "left-edge")
    )
  ),
  edge_groups = list(
    edge_group(id = "left-edge", position = "left", initial_size = 220)
  )
)
```

| Surface | |
| --- | --- |
| `edge_group()` | build one |
| `dock_view(edge_groups = )` | place them at init, before panels, so a panel can name one |
| `add_edge_group()` / `remove_edge_group()` / `set_edge_group_visible()` | drive them from the server |
| `get_edge_groups()` / `is_edge_group_visible()` | read the state back |

This supersedes #66. That branch was built on the v6-era `chore-dockview-v7` branch, which is now 51 commits behind, and it predates both the v8 serialisation of edge groups and v8's module licensing split. The R surface it designed carries over almost unchanged; what follows is what v8 made different.

## Three things found by running it, not by reading it

I probed a live dock before writing the R helpers rather than inferring the shape from the TypeScript, because `SerializedEdgeGroup` declares its group node as `unknown`. That turned up all three of these.

**Edge groups serialise beside the grid, so the group-level helpers were blind to them.** A rail serialises under a separate `edgeGroups` key, one entry per edge, each holding `size`, `visible` and a `group` node with `views` / `activeView` / `id`. Nothing appears in `grid.root`. The observable consequence, confirmed on a live dock: a rail's panels showed up in `get_panels_ids()` but in none of `get_groups_ids()`, `get_groups_panels()` or `get_active_views()`, and because `activeGroup` *is* the rail's id when it holds focus, `get_active_panel()` returned `NULL` rather than the focused panel. Those four helpers now fold edge groups in, so `get_groups_ids()` returns `c("1", "left-edge")` and `get_active_panel()` returns `"tree"`.

Worth noting that `get_active_panel()` degraded to `NULL` rather than erroring only because of the membership guard added in #93. Without it this would have been a `subscript out of bounds`.

**`restore_dock()` was silently diverging from the saved layout.** The round-trip test failed on its first run. dockview's `fromJSON` creates the edge groups named in the incoming state but leaves any other edge in place, so a rail added *after* a save survived a restore that never knew about it. `restoreDock()` now drops edge groups the incoming state does not name, before calling `fromJSON`. Without this the serialisation guarantee documented in `edge_group()` would have been false.

**`dock_view()` was about to churn every snapshot in the package.** Carrying `edgeGroups` unconditionally put `"edgeGroups": []` into every widget payload, including docks using no rails, which broke unrelated `_snaps` files. The key is now only carried when there is something to carry.

## Licensing

Click-to-collapse, drag-to-reveal and the pinnable tool-window behaviour are implemented by modules that ship in the commercially licensed `dockview-enterprise`, which this package does not bundle. `auto_hide` therefore gets no named argument, since it would be inert; it stays reachable through `...` for anyone who installs the module, and the roxygen says so.

An edge group here is a static rail: resize it with the sash, or collapse it from the server with `set_edge_group_visible()`. That is why `is_edge_group_visible()` exists, to read that state back.

## Verification

- Full suite green, 7 files plus the new one, with no snapshot churn outside `_snaps/edge_group/`.
- `R CMD check`: 0 errors, 0 warnings, 2 NOTEs, both pre-existing and environmental (unverifiable time, and a local gitignored file at top level).
- New tests cover the constructor and its validation, `check_edge_groups()`, `dock_view()` forwarding and rejection, each of the three proxies including their error paths, an app test asserting the folded-in helpers across add / hide / show / remove, and a save-and-restore round-trip that asserts the rail comes back and the rail added after the save does not.

## Not included

#66 also carried `normalize_dock_dims()`, a snapshot-stability helper unrelated to edge groups and unused by these tests.

Closes #66 once merged, though #66 should be closed by its author rather than by this PR.


christophsax:
Upvote 👍 I think these are great for assistant, workflow, and other 'global' stuff.

srcjs/modules/dock.js
```diff
@@ -41,10 +41,19 @@ const instantiateDock = (id, x) => {
   }))
 }
 
+// Edge groups are created before the panels so a panel can name one in
+// `position.referenceGroup` and land in the rail on the first pass.
+const initEdgeGroups = (x, api) => {
```
nbenn: Every edge group created through this path loses its `initial_size`. At 1211x752, `initial_size = 220`, `initial_size = 400`, and no `initial_size` at all each serialise as `size: 100.666666666667`. The same `edge_group(initial_size = 300)` through `add_edge_group()`, once the dock has settled, serialises as `300`.

The two paths differ in geometry. This function runs inside `renderValue`, which the comment in `srcjs/widgets/dockview.js` describes as the point where the grid is still zero-sized, so the splitview has no space to honour the request against and the later relayout distributes it away. That mechanism is inference; the numbers above are not.

The init path is the one the roxygen, the PR summary and `inst/examples/edge_groups/app.R` lead with, so `initial_size = 220` produces a rail around 100px wide. The proxy path, which does honour it, is not demonstrated anywhere.

R/edge_group.R
```diff
@@ -0,0 +1,159 @@
...
+  id <- as.character(id)
+
+  if (!is.logical(collapsed) || length(collapsed) != 1) {
+    stop(sprintf(
+      "<EdgeGroup (ID: %s)>: `collapsed` must be a single boolean value.",
+      id
+    ))
+  }
+
+  options <- dropNulls(list(
```
nbenn, line 102: The `collapsed` argument gets a precise error for a wrong type, while `initial_size`, `minimum_size`, `maximum_size` and `collapsed_size` go through `dropNulls()` unchecked. Passing `initial_size = "220"` or `minimum_size = -5` is accepted and forwarded verbatim.

The values themselves are not checkable: `initialSize` is a request against a splitview whose available space is unknown at construction, and the min/max are constraints dockview resolves at layout. Type and domain are — single value, numeric, finite, non-negative, and `minimum_size <= maximum_size`.

Two more in the `id` guard on line 60. Calling `edge_group(id = c("a", "b"))` gives `'length = 2' in coercion to 'logical(1)'` out of `||` rather than the intended message. And `edge_group(id = NA)` passes, since `nzchar(NA_character_)` is `TRUE`; it serialises as `{"id": null, ...}`, and `createGroup` then assigns its own (`if (!id) id = this.nextGroupId.next()`).

R/utils.R
```diff
@@ -149,6 +149,15 @@ validate_position_names <- function(id, position) {
 #' @keywords internal
 validate_position_direction <- function(id, position) {
   direction <- position[["direction"]]
+
+  # `direction` is optional when targeting a `referenceGroup`: the panel joins
+  # that group with no further placement, which is how a panel is put into an
+  # edge group. It stays required for a `referencePanel`, where the direction is
+  # the whole instruction.
+  if (is.null(direction) && !is.null(position[["referenceGroup"]])) {
```
nbenn, line 164: Making `direction` optional here enables `position = list(referenceGroup = "left-edge", index = 0)`, which `process_panel_position()` then warns about: `` `index` is ignored unless `direction` is "within" ``. The index is not ignored. Two panels sent to the same rail with the second at `index = 0` come back in tab order `b a`, and the `isPanelOptionsWithGroup` branch of `_doAddPanel` reads `index = options.position.index` with no reference to `direction`.

The check on lines 111-116 predates this, from when `direction` was mandatory and `referenceGroup` plus `index` could not occur together.

R/edge_group.R
```diff
@@ -0,0 +1,159 @@
...
+#' @param ... Other options forwarded to `api.addEdgeGroup()`. See
+#'   \url{https://dockview.dev/docs/core/groups/edgeGroups}. Use the
+#'   JavaScript option names (camelCase) for these extras.
+#'
+#'   Two of those options need a dockview module that this package does not
+#'   bundle, and have no effect here: `autoHide`, the pinnable tool-window
+#'   behaviour, and anything relying on `dockToEdgeGroups`, which reveals an
+#'   edge by dragging onto it. Both ship in the commercially licensed
+#'   `dockview-enterprise`. Setting either logs an error in the browser console
+#'   naming the missing module. Click-to-collapse is part of `autoHide`, so an
```
nbenn: Click-to-collapse ships in the MIT core.

- On a rail holding one panel, with no enterprise module loaded, clicking the active tab takes it from 94px to 44px with `isCollapsed()` true, and clicking again returns it to 94px and false.
- The bundle's `ENTERPRISE_MODULE_NAMES` set is `AdvancedOverflow`, `AutoEdgeGroup`, `AutoHideEdgeGroup`, `ContextMenu`, `DndCompass`, `KeyboardDocking`, `KeyboardNavigation`, `LayoutHistory`, `License`, `MultiRowTabs`, `PinnedTabs`, `SmartGuides`. No `EdgeGroup`. And `OPTION_MODULE_RULES` gates two edge-group options, `autoHideEdgeGroups` on `AutoHideEdgeGroup` and `dockToEdgeGroups` on `AutoEdgeGroup`.
- Upstream's edge-group page documents "clicking the active tab toggles collapse/expand" as core, marking only auto-hide and dock-to-edge-groups as enterprise.

So a rail is not static. And `set_edge_group_visible()` is a different operation: it renders 0px with `visible: FALSE` and leaves `collapsed` untouched, where a collapse leaves the 44px strip standing. The `collapsed` parameter on line 22 is core `EdgeGroupOptions`, and `collapse()` / `expand()` / `isCollapsed()` are public on `DockviewGroupPanelApi` through `api.getEdgeGroup(position)`.

The paragraph also stands in `R/proxy.R`, `NEWS.md` and the PR summary. Restricting it to `autoHide` and `dockToEdgeGroups` would match what the module list gates.

tests/testthat/test-edge_group.R
```diff
@@ -0,0 +1,247 @@
...
+  expect_setequal(app$get_value(export = "panels_ids"), c("main", "tree"))
+  expect_identical(
+    app$get_value(export = "groups_panels")[["left-edge"]],
+    list("tree")
+  )
+  expect_identical(app$get_value(export = "edge_positions"), "left")
+  expect_true(app$get_value(export = "left_visible"))
+  # No right-hand rail yet, so visibility is unknown rather than FALSE.
+  expect_null(app$get_value(export = "right_visible"))
+  app$expect_values(input = FALSE, output = FALSE, export = TRUE)
```
nbenn, line 188: The PR summary leads with `get_active_panel()` returning `NULL` rather than the focused panel, and no assertion here reads it. The app exports `active_panel` and the snapshot records `"active_panel": "tree"`, so it is covered inside a blob that `snapshot_accept()` can rewrite. The sibling asserts it directly, `expect_identical(active_panel, "3")` in `test-dock_state.R:51-52`.

Also `get_active_views()` is one of the four helpers this diff changes, and the example app never exports it, so its edge-group folding is unasserted; `test-dock_state.R:54-55` covers the grid case. Its `NA_character_` branch is reachable here, since the state after `app$click("add")` holds a rail with a `group` node and no `activeView`.

One line each: `expect_identical(app$get_value(export = "active_panel"), "tree")`, plus an `active_views` export to assert against.

nbenn (commented):
Restore handling: upstream's `_doClear()` skips edge groups, emptying their panels and continuing rather than removing them, and `deserializeEdgeGroups()` only adds when the position is free and never removes. A rail added after a save survives `fromJSON`, so the pre-removal loop in `restoreDock()` is doing real work, behind `setRestoring()`. The `visible` field is required on `SerializedEdgeGroup`, so the `isTRUE()` read is safe. Suite green locally, CI green on five platforms.

**The `initial_size` argument has no effect through `dock_view(edge_groups = )`.** At 1211x752, `220`, `400` and unset all serialise as `size: 100.666666666667`. The same value through `add_edge_group()`, once the dock has settled, comes back exact. That is the path the roxygen, the summary above and the example app lead with, so the documented call yields a rail under half the width it asks for.

**Click-to-collapse is not gated.** Clicking a rail's active tab collapses it to 44px, and clicking again expands it, with no enterprise module loaded. The bundle's `ENTERPRISE_MODULE_NAMES` set has no `EdgeGroup`, `OPTION_MODULE_RULES` gates only `autoHideEdgeGroups` and `dockToEdgeGroups`, and upstream documents the active-tab toggle as core. The claim stands in `R/edge_group.R`, `R/proxy.R`, `NEWS.md` and the summary above. The `NEWS.md` copy is the one that ships.

Four smaller things inline: validation on the size options, a warning that fires on a spelling this PR legalises, two missing assertions, one dead `Filter`.

Two upstream oddities. Dockview's docs say "adding a new panel to it will expand it again", yet a rail that auto-collapsed on being emptied stayed at 44px after `add_panel()` refilled it, expanding only on a tab click. And `evalDockView` reports `Error in http: ...` in dev mode, because its stack regex matches the URL rather than a function name. Both predate this branch.

R/dock_state.R
```diff
@@ -69,6 +69,56 @@ get_grid <- function(dock) {
...
+  validate_edge_position(position)
+  edge <- get_edge_groups(dock)[[position]]
+  if (is.null(edge)) {
+    return(NULL)
+  }
+  isTRUE(edge[["visible"]])
+}
+
+#' @keywords internal
+# The group node of each edge group, named by group id. An edge that has never
```
nbenn: The `Filter` here cannot drop anything. Upstream sets `entry.group = group.toJSON()` for every live edge group, so a `group` node carrying an `id` is always present — an empty rail serialises as `group: {views: [], id: "right-edge", headerPosition: "right"}`. The only entry lacking one is deleted whole, in the `isAutoReveal(group) && group.model.isEmpty` branch, which needs the enterprise drag-reveal to reach. With the `Filter` gone the second length guard is dead too, and the comment above goes with them.

That leaves:

```r
edge_group_nodes <- function(dock) {
  groups <- lapply(get_edge_groups(dock), `[[`, "group")
  if (length(groups) == 0) {
    return(list())
  }

  setNames(groups, vapply(groups, `[[`, character(1), "id"))
}
```

Identical to the current version on `NULL`, `list()`, one rail, two rails with one empty, and all four edges occupied. The remaining guard has to stay: `setNames(list(), character(0))` carries a names attribute, which would turn `get_groups_ids()` from `NULL` into `character(0)` on an empty dock.

DivadNojnarg:
@nbenn All seven confirmed, and all fixed in 45dbd50. Two of them were mine to own rather than tidy up, so taking those first.

## `initial_size`

Confirmed and fixed. Reproduced your A/B on the existing example: `220` through `dock_view(edge_groups = )` came back as `155`, the same `220` through `add_edge_group()` as `218.33`. Both now land on `220` exactly.

Your inferred mechanism was right, with one detail that turned out to matter. The container *does* have its real size at that point (measured 1181x752 at `renderValue`); it is dockview that has not laid out yet, so the splitview had no space to take 220px from and the relayout distributed it proportionally. So the fix is simply to lay the grid out against the container before creating the rails:

```js
const container = document.getElementById(id);
if (container && container.clientWidth > 0 && container.clientHeight > 0) {
  api.layout(container.clientWidth, container.clientHeight);
}
```

That keeps creation before `initDockPanels`, so `referenceGroup` still resolves on the first pass. The early return for "no edge groups" sits ahead of it, so a dock without rails never reaches the call and no other snapshot moved. A container measuring 0x0, a dock in a hidden tab, skips it and gets laid out by the seed as before.

Worth noting I went looking for the other fix first, re-applying sizes after the seed, and it is not available: there is no public edge-group resize. `getEdgeGroup(position).setSize({width: 320})` is inert, `setSize` only fires `_onDidSizeChange` and nothing on the shell listens for an edge view. The only path to `_shellManager.resizeView` is `setEdgeGroupCollapsed`, via `lastExpandedSize`. So laying out first is not just the tidier option, it is the one that works without reaching into internals.

## Click-to-collapse

You are right and the claim was wrong. I verified it three ways rather than take it from either direction this time: `ENTERPRISE_MODULE_NAMES` is the twelve you list with no `EdgeGroup`; `OPTION_MODULE_RULES` gates `autoHideEdgeGroups` on `AutoHideEdgeGroup` and `dockToEdgeGroups` on `AutoEdgeGroup` and nothing else edge-related; and the toggle is in the core bundle gated only on placement:

```js
tab.onTabClick((event) => {
  if (event.defaultPrevented) return;
  if (this.group.api.location.type !== "edge") return;
  if (this.group.activePanel === panel) if (this.group.api.isCollapsed()) this.group.api.expand();
  else this.group.api.collapse();
  ...
```

For the record on how it got in: it came from a review comment on the v8 upgrade reporting `pinEdgeGroup` / `autoHideEdgeGroup` / `peekEdgeGroup` all logging the missing-module error. Those three are genuinely enterprise, but generalising from them to the plain tab toggle was wrong, and I repeated it into the roxygen, `R/proxy.R`, `NEWS.md` and the summary instead of checking the module list. "A static rail" is gone from all four. The enterprise note is now limited to `autoHide` and `dockToEdgeGroups`, and your collapse-versus-visible distinction is documented, since the two really are different operations.

## The other five

- **Size validation.** All four options now type and domain checked, single finite non-negative, plus `minimum_size <= maximum_size`. Agreed the values themselves are not checkable here. `initial_size = "220"` and `minimum_size = -5` now error.
- **`id` guard.** Fixed both. `length != 1` and `NA` are rejected before the `||` can coerce, so `c("a", "b")` gets the intended message rather than `'length = 2' in coercion to 'logical(1)'`.
- **The `index` warning.** Correct, and it was my change that legalised the combination. The warning is now limited to placements that create a new group, so `referenceGroup` plus `index` passes quietly. Thanks for checking the tab order rather than assuming from the message.
- **The dead `Filter`.** Confirmed at `dockview-core.js:16252`: `entry.group = group.toJSON()` runs for every live edge group, and the only omitting path `delete`s the whole entry. Taken as written, including keeping the length guard for the `setNames(list(), character(0))` names attribute, which I would have removed along with the rest.
- **The two assertions.** Both added. `active_panel` and `active_group` are now asserted directly, and the example exports `active_views`. I also used your point about the empty rail to cover the no-`activeView` branch explicitly: after `add`, the new rail must be absent from `get_active_views()` rather than present as `NA`.

## The two upstream oddities

Agreed both predate this branch and neither is in scope here. The `evalDockView` stack regex matching a URL is worth its own issue since it makes every dev-mode error message useless, not just the edge-group ones. Happy to file it unless you already have.

Suite green locally, CI running.

